PR #84, opened 2014-02-09
Feature for field method, passing a type option will accepts any active record sql types and type casts the value when it is  set.

zeisler:
This feature is helpful because I use ActiveHash for mock objects in my tests and this would allow them to more resemble ActiveRecord's functionality.


kbrock:
Virtus has been discontinued and this diverted a bit.

If you still want this approach, please rebase and reopen. thanks

PR #91, opened 2014-06-25
support for find(:first, ...)

Nowaker:
```
Model.find :first, :conditions => {:id => 1}
Model.find 1
```

Expected: both of these return the same
Actual: ActiveHash::RecordNotFound (Couldn't find User with ID=first).

This fixes it.

Note: I didn't break Rails 4 tests. They didn't work in the first place.

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -183,6 +183,8 @@ module ActiveHash
             nil
           when :all
             all
+          when :first
+            all(*args).first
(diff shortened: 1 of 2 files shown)


kbrock:
@nowalker do you still want this functionality in? Or have you moved over to the newer syntax?

Nowaker:
Both. I guess it's okay to target multiple versions because people use multiple versions.

PR #93, opened 2014-08-17
proposed feature: support for Yaml symbol keys

ultrasaurus:
I recently tried to switch from using ActiveHash to ActiveYaml
ActiveHash supports both symbol and string keys
this change allows for indifferent_access option 
this PR is on top of pr_90 branch where tests pass

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ rvm:
   - "1.8.7"
   - "1.9.3"
   - "2.1.0"
+  - "2.1.1"
   - jruby
   #- rbx-2.2.3
 gemfile:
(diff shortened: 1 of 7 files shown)


zilkey:
Fancy :)

Any reason to not use ActiveSupport's HashWithIndifferentAccess?  http://api.rubyonrails.org/v4.2/classes/ActiveSupport/HashWithIndifferentAccess.html

ultrasaurus:
@zilkey wow.  This was a while ago!  Sorry if I'm missing something obvious... how would HashWithIndifferentAccess solve this?  do you mean somehow use that instead of ActiveYAML?  or somewhere in the implementation?

syguer:
@ultrasaurus while ago again.
It means you should use `HashWithIndifferentAccess` instead of your implementation(`#deep_symbolize_keys`)

Is you still interested in this?

syguer:
I'll close this. If you need this again, please reopen.
Thanks.

PR #97, opened 2014-09-26
add default value as lambda option in active hash

wogg:
Right now the way defaults work is that the value passed to default is initialized at class level.
This works fine for values but does not work for objects like hashes.

I have made it conditional to test if default assignment is a lambda, and calls / assigns field if it is. This allows you to declare a Hash as the default value, with that Hash being scoped in instance rather than class (or any object as default).

I haven't looked at this behavior in modules other than the ActiveHash one.

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -269,7 +269,8 @@ module ActiveHash
       def define_getter_method(field, default_value)
         unless has_instance_method?(field)
           define_method(field) do
-            attributes[field].nil? ? default_value : attributes[field]
+            
+            attributes[field].nil? ? (default_value.is_a?(Proc) ? (attributes[field]= default_value.call) : default_value) : attributes[field]
           end
         end
       end


kbrock:
Supporting a proc looks good.
Seems odd that using a proc for a default value changes the stored value in the originating record, while the other types do not.

Specs would be nice.

Anyone have an opinion on whether this should modify the base records?

kbrock:
I'm thinking:

- creating the records should set the value from the default not the attr getter.
- have a sweep after loading a yaml file to change all the values

But want a quick yea/nea from someone else like @zilkey

zilkey:
@kbrock / @wogg I'm having a hard time seeing why we'd want to do this with default values, as opposed to asking developers to just manage their instances on their own.

Since it's a getter, I would expect a Proc default value to work like `Hash.fetch`, where you either get the stored value, or it runs the proc anew every time.  I don't have a lot of other patterns to compare this to that share the same behavior of:

- If you set a value, use it and store it
- If you don't set a value, use a default value
- If that value is a Proc, call the proc _the first time_ and then return the cached value thereafter

I think in this case developers should just construct their instances with the exact objects they want.

Thoughts?

kbrock:
@zilkey I think I was reacting to #110 that wanted to search to work better with the default.

I was thinking that if the core record contained the correct value to begin with, then a lot of specialized search logic could be avoided.

Think I keep going back to your statement that this is basically static data and could just have the intended values that the developer wants in the first place rather than have a bunch of manipulation. (not sure if I misinterpreted your view)

kbrock:
>I think in this case developers should just construct their instances with the exact objects they want.

Can we close this?
I don't feel comfortable mutating the original object

syguer:
> I don't feel comfortable mutating the original object

I agree with you

PR #103, opened 2015-06-17
exclude tests on rails 3.1 and 3.2 with ruby 1.8.7

corrupt952:
```
Gem::InstallError: i18n requires Ruby version >= 1.9.3.
Installing tzinfo 0.3.44
An error occurred while installing i18n (0.7.0), and Bundler cannot
continue.
Make sure that `gem install i18n -v '0.7.0'` succeeds before bundling.
```


corrupt952:
It isn't good enough.

flavorjones:
I just fixed this on master.

PR #104, opened 2015-06-17
Unsupport ruby 1.8.7 & fix no method error

corrupt952:
[i18n](https://github.com/svenfuchs/i18n) supported rubies is 1.9.3 or greater.


flavorjones:
I believe this has been fixed on master. We should plan for deprecating 1.8.7, but it should be something we communicate well and do in a minor or major version bump, not a point release.

PR #111, opened 2015-10-29
Improvement `ActiveHash::Base#find_by`

i2bskn:
Use `find_by_id` to search by `id` and omitted unnecessary loop. (`find` is using `find_by_id`)
Has become possible to search by id of string.

```
# Country.data = [{:id => 1, :name => "US"}]
Country.find("1") #=> return record
Country.find_by(id: "1") #=> before returns nil, after returns record
```

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -157,13 +157,28 @@ module ActiveHash
       def where(options)
         return @records if options.nil?
         (@records || []).select do |record|
-          options.all? { |col, match| record[col] == match }
+          match_options?(record, options)
         end
       end
 
       def find_by(options)
-        where(options).first
+        return all.first if options.nil?
+        options.symbolize_keys!
+
+        if id = options.delete(:id)
+          candidates = Array(id).map { |i| find_by_id(id) }
+        end
+
+        (candidates || all).detect do |record|
+          match_options?(record, options)
+        end
       end
+
+      def match_options?(record, options)
+        options.all? { |col, match| record[col] == match }
+      end
+
+      private :match_options?
 
       def count
         all.length
--- a/spec/active_hash/base_spec.rb
+++ b/spec/active_hash/base_spec.rb
@@ -292,6 +292,27 @@ describe ActiveHash, "Base" do
       record.id.should == 1
       record.name.should == 'US'
     end
+
+    it "finds the record with the specified id as a string" do
+      record = Country.find_by(:id => 1)
+      record.name.should == 'US'
+    end
+  end
+
+  describe ".match_options?" do
+    before do
+      Country.data = [
+        {:id => 1, :name => "US"}
+      ]
+    end
+
+    it "returns true when matched options" do
+      Country.send(:match_options?, Country.first, {:name => "US"}).should be_truthy
+    end
+
+    it "returns false when unmatched options" do
+      Country.send(:match_options?, Country.first, {:name => "Canada"}).should be_falsey
+    end
   end
 
   describe ".count" do
--- a/spec/active_hash/base_spec.rb
+++ b/spec/active_hash/base_spec.rb
@@ -294,24 +294,16 @@ describe ActiveHash, "Base" do
     end
 
     it "finds the record with the specified id as a string" do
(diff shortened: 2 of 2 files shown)


spec/active_hash/base_spec.rb
```diff
@@ -292,6 +292,27 @@ class Region < ActiveHash::Base
...
       record.name.should == 'US'
     end
+
+    it "finds the record with the specified id as a string" do
+      record = Country.find_by(:id => 1)
+      record.name.should == 'US'
+    end
+  end
+
+  describe ".match_options?" do
```
kbrock: since `.match_options?` is a private method, you may want to implement this test from `find_by`.

That way if the internal implementation changes (and the external implementation stays the same), the tests won't have to change.

kbrock:
This looks great

i2bskn:
Thanks for your reply. I removed test of private method.
(and fixed id value to string from int)

kbrock:
@i2bskn oh noes. Those tests were great - just send them through the public.

Can you `reflog` those tests and put them in, just going through the public interface?

If not, then it is all good.

i2bskn:
@kbrock Sorry for misunderstanding. Is this alright? (put the test to `.find_by` block)
Or `match_options?` execute through `ActiveHash::Base.find_by`? (e.g. `Country.find_by(:name => "US").should_not be_nil`)

spec/active_hash/base_spec.rb
```diff
@@ -292,6 +292,19 @@ class Region < ActiveHash::Base
...
       record.name.should == 'US'
     end
+
+    it "finds the record with the specified id as a string" do
+      record = Country.find_by(:id => '1')
+      record.name.should == 'US'
+    end
+
+    it "returns true when matched options" do
+      Country.send(:match_options?, Country.first, {:name => "US"}).should be_truthy
```
kbrock:
```ruby
expect(Country.find_by(:name => "US").it).to eq([1]) # or what ever you think would be better
```

spec/active_hash/base_spec.rb
```diff
@@ -292,6 +292,19 @@ class Region < ActiveHash::Base
...
+    end
+
+    it "returns false when unmatched options" do
+      Country.send(:match_options?, Country.first, {:name => "Canada"}).should be_falsey
```
kbrock:
```ruby
expect(Country.find_by(:name => "Canada").it).to be_empty
```

kbrock:
You will still have the tests, but it will be testing against a public interface.

This is looking great

i2bskn:
I was fixed it :bow:

kbrock:
@zilkey any thoughts before I merge?

lib/active_hash/base.rb
```diff
@@ -157,14 +157,29 @@ def all(options={})
...
         return @records if options.nil?
         (@records || []).select do |record|
-          options.all? { |col, match| record[col] == match }
+          match_options?(record, options)
         end
       end
 
       def find_by(options)
-        where(options).first
+        return all.first if options.nil?
```
zilkey: When calling `find_by` with no arguments, this will act like `first`, which seems strange.  Previously it would return `nil`.  Can you just return `nil` instead of returning `all.first`?

kbrock: Looks like Active record acts like `first`.

i2bskn: `find_by` looks to specification to return the `first` when arguments is `nil`.
https://github.com/zilkey/active_hash/blob/master/spec/active_hash/base_spec.rb#L268

kbrock:
@zilkey Not sure the right course of action here. If you look at how rails defines [find_by](http://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find_by), `all.first` makes sense.

```ruby
def find_by(*args)
  where(*args).take
rescue RangeError
  nil
end
```

Seems that when in doubt, defer to rails. But it is tricky when this changes the current way an api works.

kbrock:
closing this - prefer #117 (which has some of this PR in it)

PR #113, opened 2015-12-17
yield block on initilize

marcgreenstock:
[ActiveRecord#initialize](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/core.rb#L307) yields a block on initialize. It would be handy for ActiveHash to do the same, don't you think?

### DIFF
--- a/spec/active_hash/base_spec.rb
+++ b/spec/active_hash/base_spec.rb
@@ -15,6 +15,25 @@ describe ActiveHash, "Base" do
     expect { Country.transaction }.to raise_error(LocalJumpError)
   end
 
+  describe ".new" do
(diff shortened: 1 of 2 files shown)


spec/active_hash/base_spec.rb
```diff
@@ -15,6 +15,25 @@ class Country < ActiveHash::Base
...
+        Country.new do |country|
+          country.name = 'Germany'
+        end
+      end
+
+      it "sets assigns the fields" do
+        expect(subject.name).to eq('Germany')
+      end
+    end
+  end
```
ppworks: :cool:

PR #115, opened 2016-01-12
Support read_attribute method

ppworks:
Some gems([like a enumerize](https://github.com/brainspec/enumerize/blob/master/lib/enumerize/attribute.rb#L87-L89)) expects #read_attribute method.

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -422,6 +422,7 @@ module ActiveHash
     def _read_attribute(key)
       attributes[key]
     end
+    alias :read_attribute :_read_attribute
 
     def []=(key, val)
       attributes[key] = val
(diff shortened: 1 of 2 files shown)


kbrock:
@zilkey would it make sense to just rename `_read_attribute` to `read_attribute`?

kbrock:
It looks like defining the method should pick that up?

Won't [enumerize](https://github.com/brainspec/enumerize/blob/master/lib/enumerize/attribute.rb#L85-L86) pick up a method that has that defined?

kbrock:
Also, would it make sense to use `alias_method` instead?

Think that behaves better when there are methods defined in subclasses. And this stuff is meant to be subclassed.

lib/active_hash/base.rb
```diff
@@ -422,6 +422,7 @@ def [](key)
     def _read_attribute(key)
       attributes[key]
     end
+    alias_method :read_attribute :_read_attribute
```
ppworks: I used `alias_method` .

zilkey:
Looks good to me!

PR #117, opened 2016-02-03
Support for collections in where statement

### LINKED ISSUE #99: Support for collections in where statement
```ruby
  self.data = [
    {
      id: 1, 
      name: 'Belarus',
      country_code: 'BY'
    },
    {
      id: 3,
      name: 'United States',
      country_code: 'US'
    }
  ]
end
```
Currently it is possible to pass single country code when searching for a country. For example,
```ruby
Country.where country_code: 'BY'
=> #<Country:0x007f94ba2d22a0 @attributes={:id=>1, :name=>"Belarus", :two_letter_code=>"GB"}>
```
It would be great to pass pass an array of country codes when searching for countries, like
```ruby
Country.where country_code: ['BY', 'US']
```

kbrock:
I extended #111 to better support in clauses.

Fixes #99

### DIFF
--- a/spec/active_hash/base_spec.rb
+++ b/spec/active_hash/base_spec.rb
@@ -311,6 +311,19 @@ describe ActiveHash, "Base" do
       record.id.should == 1
       record.name.should == 'US'
     end
+
+    it "finds the record with the specified id as a string" do
+      record = Country.find_by(:id => '1')
+      record.name.should == 'US'
+    end
+
+    it "returns the record that matches options" do
+      expect(Country.find_by(:name => "US").id).to eq(1)
+    end
+
+    it "returns nil when not matched in candidates" do
+      expect(Country.find_by(:name => "UK")).to be_nil
+    end
   end
 
   describe ".count" do
--- a/spec/active_hash/base_spec.rb
+++ b/spec/active_hash/base_spec.rb
@@ -226,6 +226,10 @@ describe ActiveHash, "Base" do
       Country.where(nil).should == Country.all
     end
 
+    it "returns all records when an empty hash" do
+      Country.where({}).should == Country.all
+    end
+
     it "returns all data as inflated objects" do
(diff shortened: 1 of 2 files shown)


kbrock:
@i2bskn is this similar enough to your changes?

Hirurg103:
Yes, @kbrock thank you very much! I love this feature :+1:

i2bskn:
@kbrock Yes, That I wanted to do it has been achieved. Thanks!

SmileZero:
@kbrock :+1: Love this so much.
Just one question. What if the keys in options is a string type( except "id" ). 
I mean the following statement. I know maybe it is not active_hash style.
```
expect(Country.where("name" => %w(US Canada)).map(&:name)).to match_array(%w(US Canada))
```

spec/active_hash/base_spec.rb
```diff
@@ -311,6 +323,19 @@ class Region < ActiveHash::Base
...
       record.name.should == 'US'
     end
+
+    it "finds the record with the specified id as a string" do
+      record = Country.find_by(:id => '1')
+      record.name.should == 'US'
+    end
+
+    it "returns the record that matches options" do
+      expect(Country.find_by(:name => "US").id).to eq(1)
```
kbrock: so the idea is to use `"name"` as the key here and have it work?

SmileZero: Yes. Does it sounds reasonable?

kbrock: looks like `find_by` and `where` and others do not work with string keys. Find by `"id"` does seem to work as a string.

Seems like a separate feature request

SmileZero: :+1: Yeah, I agree that a separate feature request is better.

tricknotes:
This PR is what I want.

@zilkey 
Could you release a new version?

PR #119, opened 2016-02-10
Fix actual_root_path implementation

drewbug:


lib/active_file/base.rb
```diff
@@ -46,7 +46,7 @@ def extension
       protected :extension
 
       def actual_root_path
-        root_path  || Dir.pwd
+        root_path  || File.dirname(File.realpath(__FILE__))
```
kbrock: Would this work? `File.new(File.dirname(__FILE__)).realpath`

I'm assuming we want to use `File.dirname(__FILE__)` over `__dir__` for ruby version compatibility. too bad.

kbrock:
>Fix actual_root_path implementation

Hello @drewbug could you share what you are trying to do here?

drewbug:
@kbrock: The README says: `By default, this class will look for a yml file named "countries.yml" in the same directory as the file.`

I was just attempting to bring the actual implementation into line with this. I was bitten by this when starting a daemon from a different working directory. My easy workaround is: `set_root_path __dir__`, but this doesn't work on Ruby versions older than 2.0

kbrock:
Thanks, that is helpful.

Using dir or filename will be the same result. The directory of this gem file, not the caller's dir. Maybe caller or caller_location may be of use?

I'll try and play with this... <lots on plate disclaimer>

drewbug:
You're right, @kbrock, I didn't even think about that. D'oh!

kbrock:
@drewbug there may be a way to look at `self.class.method(?).source_location` or something?
I'm sure the answer is simple, just have to find it

PR #121, opened 2016-04-26
Version 1.4.2

rudyyazdi:
Hi Jeff

Thanks for the awesome gem.

I have added some functionality to it to resemble ActiveRecord model even better:

1- Where like querying: `Model.where("id > 3 AND value = 4")`
2- Before_filter and after_filter
3- Support for nested models. 

I have wrote new test for all the new functionality and run all the tests against all the ruby versions and specs, I just had to specify rake version 10 in the travis.yml file because rake version 11 is not compatible with ruby <= 1.8

rake build, rake install are done too.

I wanted to push it to rubygems.org but I don't have access, my user name: rudyyazdi

Please let me know if there is anything else I have to do.

Thanks and Cheers,

 @zilkey

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+2016-04-07 (v1.4.2)
+  - add support for rails like queries: Country.where("id > 5")
+  - add an after_fil_load filter.
+
 2015-09-13 (v1.4.1)
   - fix bug where `#attributes` didn't contain default values #107
   - add support for `.find_by` and `#_read_attribute`. Thanks, @andrewfader!
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 2016-04-07 (v1.4.2)
   - add support for rails like queries: Country.where("id > 5")
-  - add an after_fil_load filter.
+  - add before_filter and after_filter.
   - add support for Associations when the target model is nested under a module.
 
 2015-09-13 (v1.4.1)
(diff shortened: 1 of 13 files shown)


zilkey:
@rudyyazdi - I'm interested to hear your use cases for the filters.  When we built this library we imagined it being quick config data that didn't really "deserve" a database table for whatever reason.  I've mostly seen it used to provide hard-coded default data, in which case you fully control the data, and there's no need to validate.

What are you doing that requires filters here?

rudyyazdi:
@zilkey thanks for the comment, I had the case that I had to import a bunch of objects from file, (products, users, purchases and transactions) and then purchases and transactions would go on after import for a while. I modeled the system with those objects talking to each other (rails like models) but I had to catch the anomalies before the objects were created :)

There are probably better ways to do it, but this model worked great for me as it would fit seamlessly in case we wanted to add a database in the future :)

If you think that's unnecessary please let me know and I can remove the filters altogether.

Cheers,

syguer:
@rudyyazdi Hello, sorry for late reply 😓 

According to https://github.com/zilkey/active_hash/pull/121#issuecomment-219635399, 2 and 3 are sadly, but unnecessary.

But I think 

> 1- Where like querying: Model.where("id > 3 AND value = 4")

seems to useful for some use cases.
If you are still interested in this feature, please separate your patch by the feature and create as another PR.
Thanks.

PR #122, opened 2016-07-04
Fix setter value, no write in volatile hash in case of default values are present

frodrigo:
attributes() can return a volatile hash when default_attributes, and then set into this has are lost

https://github.com/zilkey/active_hash/blob/master/lib/active_hash/base.rb#L432

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -90,7 +90,7 @@ module ActiveHash
 
       def insert(record)
         @records ||= []
-        record.attributes[:id] ||= next_id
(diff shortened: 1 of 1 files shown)


kbrock:
In active record, I thought setting into `attributes[]` we're lost because it is cloned

frodrigo:
With this I aim to change the behaviour of object[:foo]= not object.attributes[:foo]=, but the storage is in @attributes hash.

kbrock:
LGTM. lets see if we can get master green and then merge

syguer:

PR #123, opened 2016-07-07
gemfile for ruby 1.9.3

kbrock:
Older versions of ruby need older version of json

1. I need to figure out the build system and generating the Gemfiles
2. @zilkey 1.9.3 and 2.0.0 are not supported. Do you want to still support them?


kbrock:
prefer #126 but I can revive if needed

PR #125, opened 2016-07-10
Give objects without instance variables to users

stcorbett:
Separate the original state of the records in an ActiveHash sub-class
from the records that are being passed around in the rest of the
application

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -150,7 +150,7 @@ module ActiveHash
         if options.has_key?(:conditions)
           where(options[:conditions])
         else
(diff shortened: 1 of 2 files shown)


syguer:
@stcorbett 
I agree to fix the problem.
This patch looks resolve the problem of surface, but not fundamental problem.
It seems that  fundamental problem is the fact that `ActiveHash` instance is editable.
I'll try prohibit it on next version.

Could you wait it please?

syguer:
Discussing on https://github.com/zilkey/active_hash/issues/146 so closing this.

PR #127, opened 2016-10-05
Fix ActiveYaml::Aliases when using multiple files

Flink:
The current implementation of `ActiveYaml::Aliases` does not allow to use multiples files. This patch fixes it.


kbrock:
Hello @Flink 

Could you add a test to show this in action?

Flink:
@kbrock ah yes sorry, I’ll do that!

Flink:
@kbrock Sorry for the delay, it’s done! :)

PR #129, opened 2016-10-24
gemfile for ruby 1.9.3

kbrock:
for older ruby versions, peg `json` to a compatible version

Also make gemfiles more consistent


kbrock:
sorry to merge my own. but we need master green

PR #130, opened 2016-10-25
remove warning

tamgrosser:
saw this when running tests ...

lib/active_hash/base.rb:155: warning: instance variable @records not initialized

@kbrock


tamgrosser:
travis will fail with the same error that is on master ...

kbrock:
kicked travis

tamgrosser:
thank you 🎉

PR #132, opened 2016-11-06
WIP Warning

tamgrosser:
@grosser

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 sudo: false
+cache: bundler
 rvm:
-  #- "1.8.7" # only test rails < 4.0 (deprecated, reduced coverage)
   - "1.9.3"
   - "2.1.5"
   - "2.2.4"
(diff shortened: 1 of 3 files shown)


grosser:
somehow has_insance_method? returned false after define_method was called
leading to duplicate method definitions and warnings

instead of fixing this mystery bug I'm just removing the obsolete method

@kbrock

kbrock:
@grosser Great fixes.

I'm not comfortable with the 1.8.7. I got pushback trying to drop it in another gem. (If my name were in the url I'd feel like I had more jurisdiction to make this change)

Dropping ruby would force this to 2.0, or would 1.5 work too?

If you pull out the 1.8.7, then I'll merge right away.

Alternatively, @zilkey - you ok dropping 1.8.7? I'm mixed. (if you say go, then lets merge)

grosser:
whoever is stuck on 1.8 can always use an old version of this project ... 
in 2015 that was 1% ... so let's not optimize for the `< 1%` ... 
https://semaphoreci.com/blog/2015/12/01/ruby-version-usage-in-commercial-projects-2015-edition.html

grosser:
🎉

PR #133, opened 2016-11-11
remove redundant method

tamgrosser:
@kbrock 

saw this when looking over the last PR ... probably don't need this either ;)


kbrock:
Thanks

PR #134, opened 2016-12-06
Add support for array values in where conditions

Flink:
This adds the possibility to have arrays as values in data.
So an attribute can have a value of `[1,2,3]` for example and you’ll be able to search for it by doing `.where(attribute: 3)` or `.where(attribute: [1,3])`.
Other values are still handled correctly, in particular `nil`. So if you have data with a `nil` attribute, it will correctly returns the records too: `.where(attribute: nil)`.

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -174,10 +174,8 @@ module ActiveHash
 
       def match_options?(record, options)
         options.all? do |col, match|
-          if match.kind_of?(Array)
-            match.include?(record[col])
-          else
-            record[col] == match
+          [match].flatten.any? do |value|
+            [record[col]].flatten.include?(value)
           end
         end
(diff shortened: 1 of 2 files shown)


kbrock:
@Flink Is this how postgres arrays work for you?

Something is not quite right here, but I'm not sure why I'm confused.
Also, could you add a test for arrays? You only added a test for nils.

Flink:
Woops sorry about that. Errr no I don’t know for arrays in PG, should active_hash works as PG?
And yes I’ll add the specs for arrays :)

syguer:

syguer:
@Flink If you are still interested in this feature, please add some specs please?

syguer:
I'll close this. If you need this again, please reopen.
Thanks.

PR #136, opened 2017-01-05
Don't generate instance accessors for class attributes

rainhead:
I have a model with an attribute called `filename`, which was confusingly always returning nil. These changes ensure `filename`, `root_path`, `data_loaded`, and `filenames`, which are class-level concerns, don't conflict with model use.

### DIFF
--- a/lib/active_file/base.rb
+++ b/lib/active_file/base.rb
@@ -4,7 +4,7 @@ module ActiveFile
     extend ActiveFile::MultipleFiles
 
     if respond_to?(:class_attribute)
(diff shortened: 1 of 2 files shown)


rainhead:
FWIW, I don't seem to have broken any new specs.

syguer:

syguer:
hmm...
I'm going to merge this patch if spec passed.
But essentially, if accessor is necessary, we shouldn't  use class_attributes.
I'll try to refactor.
Thanks.

PR #139, opened 2017-02-09
Rescue LoadError

bryanalves:
A bare rescue won't recover from a load error.  Options parsing
sometimes misses class_name here, which can cause LoadError when the
guess for class_name is wrong.

When this happens, we would rather assume that we aren't dealing with
an ActiveHash rather than crashing

### DIFF
--- a/lib/associations/associations.rb
+++ b/lib/associations/associations.rb
@@ -8,7 +8,14 @@ module ActiveHash
         options = our_args.extract_options!
         name = our_args.shift
         options = {:class_name => name.to_s.camelize }.merge(options)
-        klass = options[:class_name].constantize rescue nil
+        klass =
+          begin
+            options[:class_name].constantize
+          rescue
+            nil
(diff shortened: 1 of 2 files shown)


lib/associations/associations.rb
```diff
@@ -8,7 +8,14 @@ def belongs_to(*args)
         options = our_args.extract_options!
         name = our_args.shift
         options = {:class_name => name.to_s.camelize }.merge(options)
-        klass = options[:class_name].constantize rescue nil
```
kbrock: would `save_constantize` work better?

bryanalves: `#safe_constantize` is still not applicable if the Autoloader is being used.

I added a test case that simulates the use of the autoloader.  The test fails on master, and passes on this branch.

syguer:

syguer:
Looks good, I'll merge.

PR #140, opened 2017-02-24
feat/thread_safe 

alChaCC:
make sure in multi-thread env still working well

### DIFF
--- a/lib/active_file/base.rb
+++ b/lib/active_file/base.rb
@@ -13,11 +13,12 @@ module ActiveFile
 
       def delete_all
         self.data_loaded = true
+        Thread.current[:data_loaded] = true
         super
       end
 
       def reload(force = false)
-        return if !self.dirty && !force && self.data_loaded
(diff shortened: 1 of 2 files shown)


syguer:
Thanks for your contribution 👍 
But it looks correct to reload data when  other thread call `detete_all`.

kbrock:
Ok, took me a while to find my issue.

First off, adding a global like this makes me nervous.

Second, this is setting a variable but not clearing it.
Will need an `ensure` to clear it.
Although, not sure what issues would happen from a `delete_all` calling another `delete_all`

Third and most significant, `ActiveHash` is about accessing reference data, not about writing an in-memory database. So to be honest, I don't even think `delete` and `delete_all` should even be in `active_hash`

syguer:
> First off, adding a global like this makes me nervous.

I agree.  And `Thread` is passed around by process so it looks danger to set value on overridable key.

> So to be honest, I don't even think delete and delete_all should even be part of active_hash

Yes, exactly!
I want to delete method for destroying data from next version... 😢

PR #141, opened 2017-02-28
Add rake version dependency

syguer:
Hi.
 `last_comment` method (called from rspec 2.99) has removed from Rake( >= 11.0) https://github.com/ruby/rake/commit/e76242ce7ef94568399a50b69bda4b723dab7c75
So this repository fails to build on Travis CI.
This PR avoids that problem 😄

### DIFF
--- a/gemfiles/rails_3.0.gemfile
+++ b/gemfiles/rails_3.0.gemfile
@@ -3,7 +3,7 @@ source 'http://rubygems.org/'
 gem 'activerecord', '~> 3.0.0'
 gem 'rspec', '~> 2.2.0'
 gem 'wwtd'
-gem 'rake'
+gem 'rake', '~> 10.0'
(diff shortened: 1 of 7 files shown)


kbrock:
So the option is to either upgrade rspec or downgrade rake?

syguer:
Exactly.
I tried upgrade rspec for latest https://github.com/zilkey/active_hash/pull/142, but it failed by another issue.
Rspec seems that does not work when loading minitest in [spec file](https://github.com/zilkey/active_hash/blob/master/spec/lint_spec.rb).
Does need this spec file? It looks working only ActiveRecord(< 4.1).
If it is unnecessary, I'll add removing commit on #142 and close this PR.

syguer:
How do you think? @kbrock @zilkey

PR #147, opened 2017-04-17
Fix a bug that returns nil instead of empty array from 575fe9d

syguer:
I'm sorry, I released v1.5.0 several days ago, but It has breaking bug.
I fixed that issue.

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -158,7 +158,8 @@ module ActiveHash
         return @records if options.blank?
 
         # use index if searching by id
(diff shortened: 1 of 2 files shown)


syguer:
@kbrock 
After your review, I'm going to add bump up commit and documentation :)

syguer:
@zilkey @kbrock no opinion? I'll push this fix soon because this is breaking change...

kbrock:
@syguer sorry - looks good.
Thanks for giving heads up, and thanks for not waiting too long on either of us, too.

syguer:
No problem, thank you for review!

PR #148, opened 2017-04-28
Add rails5.1

syguer:
Rails 5.1 has been released 🎉

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -35,7 +35,7 @@ module ActiveHash
         if Object.const_defined?(:ActiveModel)
           model_name.cache_key
         else
-          ActiveSupport::Inflector.tableize(self).downcase
+          ActiveSupport::Inflector.tableize(self.name).downcase
         end
       end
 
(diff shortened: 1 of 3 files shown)


lib/active_hash/base.rb
```diff
@@ -35,7 +35,7 @@ def cache_key
         if Object.const_defined?(:ActiveModel)
           model_name.cache_key
         else
-          ActiveSupport::Inflector.tableize(self).downcase
+          ActiveSupport::Inflector.tableize(self.name).downcase
```
syguer: I don't know why, but this code has been failed.

syguer:
@zilkey @kbrock can you check?

kbrock:
@syguer I currently have no projects that use active_hash.
I believe in the project and want to help where I can, but I have no corpus where I can run this.

Is there something in particular that I can do, or are you asking me to just kick the tires?
Keenan

syguer:
@kbrock Thanks for reply. just kick the tires please!

grosser:
need to get rid of quoted_id too see https://github.com/rails/rails/pull/27962#issuecomment-298498196
basically make the method not be defined if it's rails 5.1+

syguer:
@grosser Thanks for information. It seems to be just deprecated. Could you create an issue about your problem?

grosser:
here ya go: https://github.com/zilkey/active_hash/issues/150
should still be taken care of, deprecations are spammy :D

syguer:
@grosser thanks! :)

PR #158, opened 2018-01-16
Refactor association methods + fix issue with `#find_by_id` 

tiagopog:
## What's this PR do?

It changes the use of  `#find_by_id` in favor of `#find_by`. It also refactors some portion of the code related to the associations macros (`belongs_to`, `has_one`, `has_many`).

## TODOs

- [x] Create an abstract method to fetch associations;
- [x] Use `#find_by` rather than `#find_by_*` when fetching single associations;
- [x] Pass specs.

## Background context

Recently we went through a weird situation at [Beauty Date](https://beautydate.com) while running our seeds file. It turned out that a model based on `ActiveHash` was not retrieving a `belongs_to` association for calling the underlying `#find_by_id` on the related model and returning `nil`.

I did a little debug and figured out that, unlike the `#find_by_*`, calling the `#find_by` method was actually working as expected. I don't know yet the reason for this bug to be happening apparently 
 only when executing the seed task. Maybe it's something related to the seed context and the metaprogamming applied while defining those dynamic finders. I sure will need to make a deeper debug on it.

For now this fork has fixed the issue. Anyway, I think that avoiding unnecessary is a good practice (safer, faster etc).

## Manual test

```ruby
# belongs_to
City.belongs_to :country
City.first.country

# has_many
Country.has_many :cities
Country.first.cities

# has_one
Country.has_one :mayor
Country.first.mayor
```

### DIFF
--- a/lib/associations/associations.rb
+++ b/lib/associations/associations.rb
@@ -1,8 +1,19 @@
 module ActiveHash
   module Associations
+    def fetch_associations(target_model:, method: :find_by, filter:, conditions: nil)
+      base = apply_scope(model: target_model, conditions: conditions)
+      base.send(method, filter)
+    end
+
+    def apply_scope(model:, conditions: nil)
+      if conditions && model.respond_to?(:scoped)
+        model.scoped(conditions: conditions)
+      else
+        model
+      end
+    end
 
     module ActiveRecordExtensions
-
       def belongs_to(*args)
         our_args = args.dup
         options = our_args.extract_options!
@@ -34,7 +45,7 @@ module ActiveHash
         options[:shortcuts] = [options[:shortcuts]] unless options[:shortcuts].kind_of?(Array)
 
         define_method(association_id) do
-          options[:class_name].constantize.send("find_by_#{options[:primary_key]}", send(options[:foreign_key]))
+          options[:class_name].constantize.send(:find_by, options[:primary_key] => send(options[:foreign_key]))
         end
 
         define_method("#{association_id}=") do |new_value|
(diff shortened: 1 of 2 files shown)


lib/associations/associations.rb
```diff
@@ -92,65 +103,72 @@ def self.included(base)
...
-          klass = options[:class_name].constantize
-          primary_key_value = send(options[:primary_key])
-          foreign_key = options[:foreign_key].to_sym
-
-          if Object.const_defined?(:ActiveRecord) && ActiveRecord.const_defined?(:Relation) && klass < ActiveRecord::Relation
-            klass.where(foreign_key => primary_key_value)
-          elsif klass.respond_to?(:scoped)
-            klass.scoped(:conditions => {foreign_key => primary_key_value})
+      def normalize_primary_key(type, association_class)
+        primary_key = type == :belongs_to ? association_class.primary_key : name.constantize.primary_key
```
serradura: @tiagopog I'd like to recommend to invoke the primary_key method just once before of `.to_sym` calling. e.g:

```ruby
klass = type == :belongs_to ? association_class : name.constantize
klass.primary_key.to_sym
```

lib/associations/associations.rb
```diff
@@ -92,65 +103,72 @@ def self.included(base)
...
+      def normalize_foreign_key(type, association_name, options)
+        foreign_key =
+          if type == :belongs_to
+            options[:foreign_key] || (options[:class_name] || association_name).to_s.foreign_key
           else
-            klass.where(foreign_key => primary_key_value)
+            name.foreign_key
           end
-        end
+        foreign_key.to_sym
```
serradura: @tiagopog similar to https://github.com/zilkey/active_hash/pull/158/files#r161891497

lib/associations/associations.rb
tiagopog: Nice catch, @serradura 👍

tiagopog:
@zilkey uh, I was wondering here whether it's really necessary to keep the support for the old/deprecated Rails' `scoped` method? As you may notice in the diff I have removed this call within the `has_many` method.

kbrock:
For reference, `scoped` was dropped in rails `4.1.0` (April 2014)

PR #160, opened 2018-02-24
add - Allow to use ERB (embedded ruby) in yml files

UgoMare:


syguer:
Thank you! Looks good 😄
Can you add documentation in this [section](https://github.com/zilkey/active_hash#activeyaml)?

UgoMare:
@syguer Sorry for late reply, I added some documentation :)

PR #163, opened 2018-04-12
Delete codes for old ActiveSupport

yujideveloper:
`Class#class_inheritable_accessor` has been deleted at ActiveSupport v4.0

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,4 +21,4 @@ platforms :ruby do
   gem 'sqlite3'
 end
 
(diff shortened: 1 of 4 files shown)


syguer:
It seems deleted from 3.2?[https://github.com/rails/rails/commit/9cafc28874a681082f9f7e1e445db91f195a25ae#diff-804c15e586f734b42be0b305082d9e46]
Anyway, it doesn't exist now. Thank you!

PR #164, opened 2018-04-27
Support pluck method

ihatov08:
Support [pluck](https://apidock.com/rails/ActiveRecord/Calculations/pluck) method.

### DIFF
--- a/spec/active_hash/base_spec.rb
+++ b/spec/active_hash/base_spec.rb
@@ -404,6 +404,23 @@ describe ActiveHash, "Base" do
     end
   end
 
+  describe ".pluck" do
+    before do
+      Country.data = [
+        {:id => 1, :name => "US"},
+        {:id => 2, :name => "Canada"}
+      ]
+    end
+
+    it "returns an two dimensional Array of attributes values" do
+      expect(Country.pluck(:id, :name)).to match_array([[1,"US"], [2, "Canada"]])
+    end
+
+    it "returns an Array of attribute values" do
+      expect(Country.pluck(:id)).to match_array([1,2])
+    end
+  end
+
   describe ".first" do
     before do
       Country.data = [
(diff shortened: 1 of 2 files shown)


syguer:
Looks good.
but I'm concern about it can be called only as ClassMethod, not as method chain. I often see that `#pluck` called as method chain like `Book.where('created_at > ?', Date.yesterday).pluck(:id)`.
Although, I know it is difficult on ActiveHash 😇

spec/active_hash/base_spec.rb
```diff
@@ -404,6 +404,19 @@ class Region < ActiveHash::Base
...
+    before do
+      Country.data = [
+        {:id => 1, :name => "US"},
+        {:id => 2, :name => "Canada"}
+      ]
+    end
+
+    it "returns an Array of attribute values" do
+      expect(Country.pluck(:id, :name)).to match_array([[1,"US"], [2, "Canada"]])
+    end
```
syguer: if calling `#pluck` with single symbol(`Country.pluck(:id)`),  it returns flatten array. right?(not like `[[1],[2]]`)
so please add the case

spec/active_hash/base_spec.rb
```diff
@@ -404,6 +404,23 @@ class Region < ActiveHash::Base
...
+      ]
+    end
+
+    it "returns an two dimensional Array of attributes values" do
+      expect(Country.pluck(:id, :name)).to match_array([[1,"US"], [2, "Canada"]])
+    end
+
+    it "returns an Array of attribute values" do
+      expect(Country.pluck(:id)).to match_array([1,2])
+    end
```
ihatov08: @syguer 
Add test for single symbol case.

syguer: great 👍

PR #168, opened 2018-11-24
Add `where` methods to the readme

hmayer00:
Document the ability to use `where` and `where.not`, which seem to work.


syguer:
Thanks 😄

PR #170, opened 2019-02-02
support i18n

Yuki-Inoue:
This PR makes ActiveHash support i18n if `I18n` is available.

With this fix, model's model_name will be configurable by locale file, for example:

```
ja:
  activemodel:
    models:
      country: "国" # Country.model_name.human # => "国"
```

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -57,11 +57,7 @@ module ActiveHash
     end
 
     if Object.const_defined?(:ActiveModel)
-      if Object.const_defined?(:I18n)
-        extend ActiveModel::Translation
-      else
-        extend ActiveModel::Naming
-      end
+      extend ActiveModel::Translation
       include ActiveModel::Conversion
     else
       def to_param
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -57,7 +57,11 @@ module ActiveHash
     end
 
     if Object.const_defined?(:ActiveModel)
-      extend ActiveModel::Naming
+      if Object.const_defined?(:I18n)
+        extend ActiveModel::Translation
+      else
+        extend ActiveModel::Naming
+      end
       include ActiveModel::Conversion
     else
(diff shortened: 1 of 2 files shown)


lib/active_hash/base.rb
```diff
@@ -57,7 +57,11 @@ def normalize(v)
     end
 
     if Object.const_defined?(:ActiveModel)
-      extend ActiveModel::Naming
+      if Object.const_defined?(:I18n)
```
syguer: [`activemodel` depends on `activesupport`](https://rubygems.org/gems/activemodel), also, [`activesupport` depends on `i18n`](https://rubygems.org/gems/activesupport).  So It looks doesn't need.
Have you ever encountered with situation that `activemodel` exists but not `i18n` ?

Yuki-Inoue: You are right, the line seems to be redundant, and I also cannot think of a situation where `ActiveModel` is defined buy I18n is not available..

I'll make it simpler.

syguer:
Thank you for your contribution. Sounds good!
Could you write test codes and write short document to README.md?
If you are busy, I'll help these job 👍

Yuki-Inoue:
@syguer Thanks for the review.  The point out you've made is fixed with the next commit.

About the document/testing, I'll try.  Document is relatively easy, whereas the test it might take some time for me to figure out how to setup... (all the locale setup usually made by rails, I find it not obvious how to write tests...)

yhirano55:
It looks nice work to me :clap: I would like to know why this pull request is not merged.

Yuki-Inoue:
From where I remember, I was asked to write tests.
I tried to do so, but had trouble; I couldn't figure out how to write test for this kind of feature.

I would be glad if anyone can help...

kbrock:
Thank you for your work.

Looks like 230 was based upon this PR, but with tests.
Closing this as we now support i18n.

Please reopen if there is still outstanding work to be done here.

PR #171, opened 2019-03-05
Allow empty YAML

ppworks:
Fix the following error occurred when YAML file is empty.

```
NoMethodError:
       undefined method `values' for false:FalseClass
     # ./lib/active_yaml/base.rb:12:in `load_file'
```


syguer:
Hi 😄 
Sorry, I pushed a commit which fixes same problem on ActiveJSON to your branch without your agreement 😋

PR #173, opened 2019-03-06
Add ::scope method (inspired by ActiveRecord)

nsommer:
This PR basically adds a `scope` class method comparable to the [interface of ActiveRecord](https://api.rubyonrails.org/classes/ActiveRecord/Scoping/Named/ClassMethods.html#method-i-scope).

Not sure how you handle contributions in general, happy to take your feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,14 @@ Country.find_all_by_name "foo"                # => returns an array of the objec
 Country.find_by_id_and_name 1, "Germany"      # => returns the first object matching that id and name
 Country.find_all_by_id_and_name 1, "Germany"  # => returns an array of objects matching that name and id
 ```
+
+Furthermore, it allows to create custom scope query methods, similar to how it's possible with ActiveRecord:
+
+```ruby
+Country.scope :english, -> { where(language: 'English') } # Creates a class method Country.english performing the given query
+Country.scope :with_language, ->(language) { where(language: language) } # Creates a class method Country.with_language(language) performing the given query
+```
+
 ## Instance Methods
 
 ActiveHash objects implement enough of the ActiveRecord api to satisfy most common needs.  For example:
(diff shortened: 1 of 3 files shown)


joshm1204:
What's the status of this PR?  I think this is useful.

nsommer:
I guess it's up to @syguer to decide.

syguer:
I'm sorry for late reply, but I'd just worried about responsibility of ActiveHash.
points are below...

1. `.scope` should not be necessary. (mostly same as define classmethod)
2. `.scope` is expected to return object what is able to chain `scoped`.[[1]](https://www.justinweiss.com/articles/should-you-use-scopes-or-class-methods/) But, you know, it can't. because ActiveHash's `.where` not does.
3. ActiveHash may behave "like" ActiveRecord, but not "completely same". [[2]](https://github.com/zilkey/active_hash/issues/80#issuecomment-35466815)

Anyway, I really respect your contribution. And I understand some people want this feature.
So finally I decide merge this PR.
Thanks

Ref:
[[1] Should You Use Scopes or Class Methods?](https://www.justinweiss.com/articles/should-you-use-scopes-or-class-methods/)
[[2] On @zilkey(Author of ActiveHash)'s comment](https://github.com/zilkey/active_hash/issues/80#issuecomment-35466815)
> ActiveHash was really intended to be used as a simple way to populate dropdowns in Rails views. If you are trying to use it as an in-memory data store, you might be disappointed at how poorly it works for that purpose.

nsommer:
Thank you so much for accepting my contribution and your remarks.

> `.scope` is expected to return object what is able to chain `scoped`.[1] But, you know, it can't. because ActiveHash's `.where` not does.

Do you think you would consider accepting a PR if I was working on making the `where`'s chainable?

syguer:
Sounds good 😄 
Before you work, let me tell you a point what is my concern.
It seems many people uses returned value as `Array`, so I think `.where` must return value as chainable and as `Enumerable` too.

nsommer:
Awesome! I'll try to find some time in the upcoming weeks to work on it.

> It seems many people uses returned value as `Array`, so I think `.where` must return value as chainable and as `Enumerable` too.

Yes, fully agree. I'll keep it in mind.

PR #175, opened 2019-06-11
`where` clause now works with range argument

okuramasafumi:
This behavior is same as ActiveRecord.
We support **endless** range here, but doesn't support **beginless** range which will be introduced in Ruby 2.7.
That's because Rails 6 supports endless range only.
https://edgeguides.rubyonrails.org/6_0_release_notes.html#active-record-notable-changes


syguer:
Looks good 👍 
Could you try to fix compatibility test using [this](https://github.com/rails/rails/pull/34906/files#diff-eeb62dea734ca99678b4576d8da5b687R645) as reference?

okuramasafumi:
@syguer I did!

syguer:
Thanks!

PR #177, opened 2019-07-20
Add ActiveHash::Base.order method inspired by ActiveRecord

Sean0628:
This PR basically tries to add `ActiveHash::Base.order` class method.
Kindly consider my request, happy to take your feedback.
Thanks.

### DIFF
--- a/lib/active_hash/relation.rb
+++ b/lib/active_hash/relation.rb
@@ -3,7 +3,7 @@ module ActiveHash
     include Enumerable
     
     delegate :each, to: :records # Make Enumerable work
-    delegate :equal?, :==, :===, :eql?, to: :records
+    delegate :equal?, :==, :===, :eql?, :sort!, to: :records
     delegate :empty?, :length, :first, :second, :third, :last, to: :records
         
     def initialize(klass, all_records, query_hash = nil)
@@ -71,7 +71,20 @@ module ActiveHash
     def reload
       @records = filter_all_records_by_query_hash
     end
-    
+
+    def order(*options)
+      check_if_method_has_arguments!(:order, options)
+      records = where({})
+      return records if options.blank?
+
+      processed_args = preprocess_order_args(options)
+      candidates = records.dup
+
+      order_by_args!(candidates, processed_args)
+
+      candidates
+    end
+
     attr_reader :query_hash, :klass, :all_records, :records_dirty
     
     private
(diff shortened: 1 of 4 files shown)


syguer:
@Sean0628
Sorry for late reply 🙇 
Your idea looks nice! but a point I was concerning long time is iwe can't use it same as `.order` on `ActiveRecod`.
Generally,  `.order` is used `.where` together. This implementation doesn't satisfied it.
But we know it was difficult on `ActiveHash` because `ActiveHash#where` doesn't return chainable value.
But today, it is fixed on #178. So I suggest 2 ways

1. (I'm afraid but) Re-implement to follow chainable
2. This PR merge to `stable-2.x` and release as `2.4.0` (After it, I will do 1)

Sean0628:
Thank you for the replay 🙇 

I am down for no1 idea.
> (I'm afraid but) Re-implement to follow chainable

To be honest, the point you mentioned was my concern too. It was real tough at the same time like you said.

However, https://github.com/zilkey/active_hash/pull/178 has been merged. So, now I am gladly to make this chainable. It might take a couple of days though.

Thank you.

Sean0628:
@syguer
Added support for `.where` chain. Would love to hear your feedback.
Thanks.

lib/active_hash/relation.rb
```diff
@@ -71,7 +71,20 @@ def pluck(*column_names)
...
+      check_if_method_has_arguments!(:order, options)
+      records = where({})
+      return records if options.blank?
+
+      processed_args = preprocess_order_args(options)
+      candidates = records.dup
+
+      order_by_args!(candidates, processed_args)
+
+      candidates
```
nsommer: This seems to returns an `Array`, but I think it would be better to return an `ActiveHash::Relation` again, so that you can chain it.

To make that work, I think instead of returning the duped & ordered `records`, writing them back to `@records` and returning `self` should do it.

Sean0628: @nsommer 
Thanks for your comment.

This method returned an `Array` object when I first implemented as you mentioned.
However, it does return `ActiveHash::Relation` after this change cf.https://github.com/zilkey/active_hash/pull/177/commits/fcfa0ab6edbbead9855f4606c3a0b8c4d99431a2.

```ruby
Country.order("code, id DESC").is_a?(ActiveHash::Relation)
=> true
Country.where(language: 'English').order(id: :desc).is_a?(ActiveHash::Relation)
=> true
```

This returns `ActiveHash::Relation` so that it can be used like following example.
```ruby
Country.order("code, id DESC").where(id: 1)
```
Does this resolve your concern?
Let me know if you need further explanation. Would love to share it.

Thanks!

nsommer: Oh, I see. I was confused because of the local `records` variable which gets a relation assigned (with `where({})`), because there is also a private method `records` which returns the data array. Maybe you can rename the local variable to something like `relation_to_order` or something like that?

Sean0628: > I was confused because of the local records variable which gets a relation assigned (with where({})), because there is also a private method records which returns the data array. 

Thank you for your feedback!
This makes a lot of sense to me. `records` was a bit confusing. 
 Renamed `records` to `relation`.

nsommer: Great 👍

syguer:
Sorry for late.
Looks good! Thank you 😄

PR #178, opened 2019-08-03
Make #where chainable

nsommer:
In #173 I committed to working on a change that makes the `#where` method chainable, just like in ActiveRecord. I often wished this was possible when using ActiveHash and it makes the `::scope` method from #173 much more useful.

Basically I introduced an `ActiveHash::Relation` class, which represents a relation/query and can be chained as often as one would like. `ActiveHash::Base` returns a relation instance when `::all` is called. All other query methods (such as `find_by` etc.) moved to `ActiveHash::Relation`. You can still call them from the base class though (they are being delegated to `::all` which as mentioned above returns a relation). `ActiveHash::Relation` also implements `Enumerable`, so you don't have to call `#to_a` explicitly. Therefore the interface of active hash classes does not really change, but is just a bit more powerful.

Please let me know what you think and what I can improve in the implementation (there is one comment about accessing the index from the relation class which is not done very nicely atm).

### DIFF
--- a/lib/active_hash/relation.rb
+++ b/lib/active_hash/relation.rb
@@ -4,9 +4,8 @@ module ActiveHash
     
     delegate :each, to: :records # Make Enumerable work
     delegate :equal?, :==, :===, :eql?, to: :records
-    
-    delegate_missing_to :records
-    
+    delegate :empty?, :length, :first, :second, :third, :last, to: :records
+        
     def initialize(klass, all_records, query_hash = nil)
       self.klass = klass
       self.all_records = all_records
--- a/lib/active_hash/relation.rb
+++ b/lib/active_hash/relation.rb
@@ -1,5 +1,5 @@
 module ActiveHash
-  class ResultSet
+  class Relation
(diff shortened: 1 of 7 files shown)


lib/active_hash/relation.rb
```diff
@@ -0,0 +1,128 @@
...
+    
+    def find_by(options)
+      where(options).first
+    end
+
+    def find_by!(options)
+      find_by(options) || (raise RecordNotFound.new("Couldn't find #{klass.name}"))
+    end
+    
+    def find(id, * args)
```
rkallensee:
```suggestion
    def find(id, *args)
```

nsommer:
@syguer any thoughts on this? (no worries, I know you're probably busy)

syguer:
Sorry for late reply 🙇 
Amazing! Looks perfect!
I get started merge this and release as v3.0!
Give me just a day!

syguer:
Finally released just now! Thanks @nsommer 😄

PR #182, opened 2019-09-29
Add #to_ary to ActiveHash::Relation

nsommer:
This PR adds `#to_ary` to `ActiveHash::Relation`, which returns a `#dup`ed array of the items of the relation. When upgrading to active_hash 3.0.0, the following line in a controller that used [active_model_serializers](https://github.com/rails-api/active_model_serializers) to return a collection of items of an ActiveHash data store did not work anymore, because active_model_serializers uses `#respond_to?(:to_ary)` to determine if an object is a collection or not (`ActiveRecord::Relation` does respond to this method). After this PR, this should work again.

```ruby
render json: MyActiveHashClass.all, each_serializer: MyActiveHashClassSerializer
```

I put the spec for this new method into a new file `spec/active_hash/relation_spec.rb`, because it would not really belong to the `base_spec.rb`.

### DIFF
--- a/lib/active_hash/relation.rb
+++ b/lib/active_hash/relation.rb
@@ -72,6 +72,10 @@ module ActiveHash
       @records = filter_all_records_by_query_hash
     end
     
+    def to_ary
+      records.dup
+    end
+    
     attr_reader :query_hash, :klass, :all_records, :records_dirty
     
     private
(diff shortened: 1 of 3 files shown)


nsommer:
@kbrock As you mentioned in my previous PR (#178), it seems like `#to_ary` also removes the necessity to call `#to_a` in `base_spec.rb`. :-)

syguer:
👍

PR #183, opened 2019-09-30
Allow #find to behave like Enumerable#find if id is nil and a block is given

nsommer:
We discovered in our codebase that we often used `#find` chained after a `#where` or `#all` query, which used to invoke `Enumerable#find` because an `Array` was returned.

Since those methods are chainable now and return an `ActiveHash::Relation`, `#find` calls invoked the implementation in the relation class.

This change makes `ActiveHash::Relation#find` compatible with both the query variant (find by id) as well as with the `Enumerable` variant (find by block criteria). This is how active record behaves, too.


ktmouk:
:+1:

syguer:
👍

PR #186, opened 2019-10-29
Remove redundant to_ary methods

Sean0628:
## Summary
- Remove redundant `#to_ary` calls after addition of `#to_ary` method to `ActiveHash::Relation` https://github.com/zilkey/active_hash/pull/182


syguer:
Thanks!

PR #189, opened 2019-11-26
Delegate :sample to `records`

1grc:
In this PR, delegate macro will accept `#sample` method to `records` ([same as ActiveRecord](https://github.com/rails/rails/blob/b13c518ddf5912c5d4fe69d508f7af04bb3b515c/activerecord/lib/active_record/relation/delegation.rb#L86))

before
```
[1] pry(main)> Foo.all.sample
NoMethodError: undefined method `sample' for #<ActiveHash::Relation:0x000055874716f4f8>

[2] pry(main)> Foo.all.to_a.sample
=> #<Foo:0x000055873fe07920 @attributes={:id=>3, :type=>"BAZ"}>
```

after
```
[1] pry(main)> Foo.all.sample
=> #<Foo:0x000055b80956ca40 @attributes={:id=>2, :type=>"BAR"}>
```

### DIFF
--- a/lib/active_hash/relation.rb
+++ b/lib/active_hash/relation.rb
@@ -5,6 +5,7 @@ module ActiveHash
     delegate :each, to: :records # Make Enumerable work
     delegate :equal?, :==, :===, :eql?, :sort!, to: :records
     delegate :empty?, :length, :first, :second, :third, :last, to: :records
+    delegate :sample, to: :records
         
     def initialize(klass, all_records, query_hash = nil)
       self.klass = klass
--- a/spec/active_hash/relation_spec.rb
+++ b/spec/active_hash/relation_spec.rb
@@ -11,6 +11,19 @@ RSpec.describe ActiveHash::Relation do
   end
   
   subject { model_class.all }
+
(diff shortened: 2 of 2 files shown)


syguer:
Thank you for your contribution 😄 
Could you write some test codes?

1grc:
> @1grc
> Thank you for your contribution 😄
> Could you write some test codes?

OK, [added test code](https://github.com/zilkey/active_hash/commit/cfbc8b64bdba833612720814987a10a380534fc7).
I worried if I should add a test code for 'return a random element or n random elements'.
Is this in line with your wishes?

syguer:
@1grc Thanks! Yes, it looks good!

[IMO]
It seems the first test is covered by the second one so I think it is enough by second case.

1grc:
@syguer Thank you for your kindness! 😄 

>[IMO]
>It seems the first test is covered by the second one so I think it is enough by second case.

Right. Do I need to delete the first test case?
(Is it unnecessary because it has already been merged?)

syguer:
No, you don't have to do. Thanks 👍

PR #195, opened 2020-01-28
Supports .pick method

yhirano55:
## Summary

Hi, I felt it's convenient to add AH#pick similar to AR#pick. So I've opened this pull request:

- Supports .pick method similar to `ActiveRecord::Relation#pick` .
- This method is provided from ActiveRecord version 6.0.

refs) `ActiveRecord::Relation#pick` implementation:

https://github.com/rails/rails/blob/6-0-stable/activerecord/lib/active_record/relation/calculations.rb#L198-L214

Thanks.

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -35,7 +35,7 @@ module ActiveHash
         filtered_records = (candidates || @records || []).reject do |record|
           match_options?(record, options)
         end
-        
+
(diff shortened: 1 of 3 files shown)


syguer:
Thanks!

yhirano55:
@syguer You, too. Thanks for your maintenance work :pray:

PR #196, opened 2020-02-08
Support JSON serialize

yhirano55:
If it use active hash with web api responses, `as_json` behavior is different between active hash's instance and active model's instance.

```
  1) ActiveHash Base #as_json returns a hash
     Failure/Error: expect(country.as_json).to eq({"id" => 1, "name" => "US", "language" => "English"})

       expected: {"id"=>1, "name"=>"US", "language"=>"English"}
            got: {"attributes"=>{"id"=>1, "name"=>"US", "language"=>"English"}}
```

We don't need `attributes` route key.

If it includes `ActiveModel::Serializers::JSON`, behavior of active hash's instance would be as same as a behavior active model's instance.

Thanks :)

---

### The reason to need this change

- When we implement web api with using active hash model and active model serializers gem, we need include `ActiveModel::Serializers::JSON` module to all active hash's models.
- It's not a little convenient for us. So I've opened this pull request.

### DIFF
--- a/spec/active_hash/base_spec.rb
+++ b/spec/active_hash/base_spec.rb
@@ -1531,4 +1531,16 @@ describe ActiveHash, "Base" do
     end
   end
 
+  describe "#as_json" do
+    before do
+      Country.data = [
+        {:id => 1, :name => "US", :language => 'English'}
+      ]
+    end
+
+    it "returns a hash" do
+      country = Country.find(1)
+      expect(country.as_json.stringify_keys).to eq({"id" => 1, "name" => "US", "language" => "English"})
+    end
+  end
 end
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -61,6 +61,7 @@ module ActiveHash
     if Object.const_defined?(:ActiveModel)
       extend ActiveModel::Naming
       include ActiveModel::Conversion
+      include ActiveModel::Serializers::JSON
     else
       def to_param
         id.present? ? id.to_s : nil


spec/active_hash/base_spec.rb
```diff
@@ -1531,4 +1531,16 @@ class Book < ActiveRecord::Base
...
+  describe "#as_json" do
+    before do
+      Country.data = [
+        {:id => 1, :name => "US", :language => 'English'}
+      ]
+    end
+
+    it "returns a hash" do
+      country = Country.find(1)
+      expect(country.as_json.stringify_keys).to eq({"id" => 1, "name" => "US", "language" => "English"})
```
yhirano55: Rails 6.0 or later returns stringified keys hash, but Rails 5.2 or earlier returns symbolized keys hash. So I used `stringify_keys` method.

pfeiffer:
This PR doesn't solvesthe underlying issue; namely that `attributes` return a hash with the keys as symbols, while `ActiveModel::Serializer` explicitly mentions that it expect it to be a string ([docs](https://api.rubyonrails.org/classes/ActiveModel/Serialization.html)).

Even with this PR, the `as_json` doesn't work as expected:

```ruby
Country.find(1).as_json
=> { id: 1, name: "US", .. }

Country.find(1).as_json(only: [:name])
=> {} # Huh!?
```

yhirano55:
@pfeiffer Thanks for the comment. Let me check it.

yhirano55:
ActiveHash's attributes are converted to symbolized keys in its constructor.

But in `ActiveModel::Serialization#serializable_hash`, it compares stringified keys and symbolised keys, so it seems it doesn't work `only` and `except` options.

https://github.com/zilkey/active_hash/blob/d2ccb223256b876a69b217ec70f314bdf71d2573/lib/active_hash/base.rb#L404-L411

ActiveModel::Serialization:

https://github.com/rails/rails/blob/ac65e560db598c7a8b1b6aaea1e5656ff3b27f08/activemodel/lib/active_model/serialization.rb#L123-L131

I consider a better solution. If you have an idea, let me know how resolve it. Thanks.

pfeiffer:
I think way to solve this is to migrate to having keys as strings in the `attributes` attribute.This is also how `ActiveModel::Attributes` and `ActiveRecord` behaves, and will allow full compatibility with `ActiveModel` mixins such as `AM::Serialization` etc.

```ruby
# ActiveRecord
class User < ApplicationRecord; end

User.find(1).attributes
# => { "id" => 1, "name": "John", ... }

# ActiveModel
class Product
  include ActiveModel::Attributes

  attribute :sku, :string
end

Product.new(sku: "red_shirt").attributes
# => { "sku": "red_shirt" }
```

It's however a breaking change, since it's keys of attribute are currently forced to be [symbols](https://github.com/zilkey/active_hash/blob/master/lib/active_hash/base.rb#L405). Personally, I think this is the way to go.

If needed, a potential migration path would be to store keys as strings with `stringify_keys!`, but still allow access to the attributes via symbol, but with a deprecation warning before dropping support for symbols as key in next major version, but this would probably require that @zilkey signs off on this migration plan.

What do you and @zilkey think?

zilkey:
What about sort of a parallel track?

- we push a version (maybe `3.1.1`?) that deprecates symbol keys, but maintains 100% compatibility (still has symbols / still has attributes)
- we push a new version `4.0.0` that makes all the breaking changes:
  - stores attributes as strings
  - returns attributes as strings
  - removes the `attributes` key from `as_json`

We'd created a 3.x.x branch here, and can release changes on both for a little while. Any reason to not do this?

pfeiffer:
> What about sort of a parallel track?

Sounds like a great plan! Happy to contribute if needed :-)

kbrock:
Looking back over this.

I think you can manually add this to each of your models to gain this functionality.

```ruby
include ActiveModel::Serializers::JSON
```

Not sure why this had not been merged.

kbrock:
kicking this build

pfeiffer:
> Looking back over this.
> 
> I think you can manually add this to each of your models to gain this functionality.
> 
> ```ruby
> include ActiveModel::Serializers::JSON
> ```
> 
> Looks like this lost steam. I'm not sure if there is desire to get this in and beak compatibility.

No the issue would still be here. The issue is that `ActiveHash` stores the attributes with symbols as keys, causing `:only` and `:except` options passed to `as_json` to not work, as it compares strings to symbols.

kbrock:
@pfeiffer ugh. you had already clearly stated the problem. sorry about that.

Are the 2 branches going to be that much different?
Do you think it will be possibly to backport the same PRs to both branches while 2 live in parallel?

kbrock:
@pfeiffer Sorry for lag on this one. (going over unresolved PRs)

So if we want this fixed we need to break compatibility and store strings as keys?
We just merged #312 which made sure all attributes were consistent - symbols for now, which seemed to be the way it was expected to work.

But looking at #310, it seems like at least `to_json` and `column_names` wants String names.

kbrock:
Rails 7.2 (maybe 7.1) has migrated to storing attributes as strings and not symbols.
Think we'll have to bite the bullet and go that route. Not sure if that is a breaking change for users with older rails versions. (or I guess all users really)

Issue #193, opened 2020-01-21
The `order` method breaks finders


bibendi:
Hi!

After updating to v3.1 to use the new `order` method, I've experienced problems with finders.

```
> CoreBy::Language.find(54)
=> #<CoreBy::Language:0x00005583fd959238 @attributes={:id=>54, :code=>"en", :title=>"English"}>

> CoreBy::Language.order(title: :asc)

> CoreBy::Language.find(54)
=> #<CoreBy::Language:0x00005583fd9c7a58 @attributes={:id=>192, :code=>"gd", :title=>"Gàidhlig"}>
```

I think there is a mutation of original records causing the breaking of `record_index`.

flavorjones:
@kbrock I think I like the fix in #261 for this (which is simpler than what's introduced in #197). WDYT about merging it?

kbrock:
Thank you for raising this issue. It is fixed by #268

PR #197, opened 2020-02-09
Fix bug that .order breaks the order of original records

ytakhs:
This PR fixes https://github.com/zilkey/active_hash/issues/193

```ruby
  field :name
  field :language
  self.data = [
    {:id => 1, :name => "US", :language => 'English'},
    {:id => 2, :name => "Canada", :language => 'English'},
    {:id => 3, :name => "Mexico", :language => 'Spanish'}
  ]
end

p Country.first.name
#=> "US"
Country.order(id: :desc)
p Country.first.name
#=> "Mexico"
```

`#filter_all_records_by_query_hash` returns all_records without copy when `query_hash` is empty, so `Array#sort!` breaks the order of original records.

Thanks.

### DIFF
--- a/lib/active_hash/relation.rb
+++ b/lib/active_hash/relation.rb
@@ -12,6 +12,7 @@ module ActiveHash
       self.all_records = all_records
       self.query_hash = query_hash
       self.records_dirty = false
+      self.order_values = []
       self
     end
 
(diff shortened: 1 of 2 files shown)


kbrock:
With this implementation, I wonder if we need `reorder` as well.

littleforest:
I'm wondering what the status of this PR is? I've been affected by #193, and have kept using version 2.3 to avoid this bug, but I am now needing to upgrade to Ruby 3. Could this PR be moved along (or closed if it is not the correct way to solve the issue?)

castlese:
@littleforest how did you end up addressing your problem? We are in the same situation right now, reliant on an old branch but need to upgrade ruby

littleforest:
@castlese I just redefined the `order` method in my model:

```
    def self.order(field)
      all.sort_by { |hash| hash[field] }
    end
```

kbrock:

PR #200, opened 2020-02-17
Add a :private field option

djberg96:
The idea here is that `data` may be coming from an external source, and you don't want it all to be publicly accessible, but you still want it to be accessible internally for other methods. So, for example, you would have something like this:

```
class Foo < ActiveHash::Base
  field :name
  field :age
  field :stuff, :private => true

  # Let's pretend this array of hashes came from an external source
   self.data = [
     {:name => "Dan", :age => 50, :stuff => "secret"},
     {:name => "Joe", :age => 35, :stuff => "also secret},
  ]

  private

  def some_internal_method
    puts "This is: " + stuff # Want stuff here
  end
end

Foo.first.stuff # should be a NoMethodError
```

~~WIP for now until I can figure it a proper implementation, and also see if such a thing is of interest.~~

Followup to https://github.com/zilkey/active_hash/issues/199


lib/active_hash/base.rb
```diff
@@ -214,17 +214,25 @@ def fields(*args)
       end
 
       def field(field_name, options = {})
+        return if already_defined?(field_name)
```
kbrock, line 217: Think this has some fun implications

lib/active_hash/base.rb
```diff
@@ -214,17 +214,25 @@ def fields(*args)
...
         validate_field(field_name)
         field_names << field_name
 
         add_default_value(field_name, options[:default]) if options[:default]
-        define_getter_method(field_name, options[:default])
```
kbrock, line 221: thinking we don't need these changes

lib/active_hash/base.rb
```diff
@@ -214,17 +214,25 @@ def fields(*args)
...
+        define_setter_method(field_name, options)
+        define_interrogator_method(field_name, options)
         define_custom_find_method(field_name)
         define_custom_find_all_method(field_name)
       end
 
+      def already_defined?(field_name)
+        field_names.include?(field_name) ||
+          self.instance_methods.include?(field_name) ||
+          self.private_instance_methods.include?(field_name)
```
kbrock, line 232:
```suggestion
          instance_methods.include?(field_name) ||
          private_instance_methods.include?(field_name)
```

test.rb
```diff
@@ -0,0 +1,29 @@
+$:.unshift 'lib'
+require 'active_hash'
+
+class Foo < ActiveHash::Base
+  field :name
+  field :age
+  field :stuff, :private => true
```
kbrock, line 7: as said above, not a fan of private stuff.

Looking forward to seeing this an actual test

lib/active_hash/base.rb
```diff
@@ -279,33 +287,36 @@ def add_default_value field_name, default_value
...
       end
 
-      def define_getter_method(field, default_value)
+      def define_getter_method(field, options)
         unless instance_methods.include?(field.to_sym)
           define_method(field) do
-            attributes[field].nil? ? default_value : attributes[field]
+            attributes[field].nil? ? options[:default] : attributes[field]
           end
+          private field if options[:private]
```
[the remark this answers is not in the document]
djberg96: > I'd prefer that if the developer manually call field to define fields, then the developer should define all fields, and it wouldn't define any fields from the hash. But if they did not call field, then the hash would define the fields.

I'm not sure what you mean by this. How would this look in practice?

test.rb
djberg96: @kbrock well, like i said, it's a case of getting data from a 3rd part that you're trying to wrap that you don't want exposed. Do you have a better suggestion?

kbrock: Hey @djberg96 - I get your use case and it sounds good.

Your example, with a few tweaks, was able to hide the attribute that you want hidden.
Since active record seems to work that way, this is the way I'm leaning.

The diff above got `test.rb` working for me.

djberg96:
@kbrock simplified as per your recommendation, and added specs.

djberg96:
@zilkey Ok, should be good to go.

lib/active_hash/base.rb
```diff
@@ -218,13 +218,20 @@ def field(field_name, options = {})
         field_names << field_name
 
         add_default_value(field_name, options[:default]) if options[:default]
-        define_getter_method(field_name, options[:default])
+        define_getter_method(field_name, options)
```
kbrock, line 221: think this is one last remnant of the previous refactoring

lib/active_hash/base.rb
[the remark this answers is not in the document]
kbrock: yea, not a change we will make now.

A developer can define fields with `field :name` or choose not to. If a developer does not define a field, it is automatically picked up.

I'd prefer if we didn't do both:

Either the developer defines fields and no fields get automatically defined.
Or the fields are not defined and it automatically discovers them.

Having it in both camps is causing this issue for you.

Again, I don't want to make a change that has potential wide spread consequences.

kbrock:
Is there a way to define a spec for this?
It would probably be your `test.rb` with minimal changes

lib/active_hash/base.rb
djberg96: I have to keep this if i want the `field :stuff, :private => true` syntax, don't I?

djberg96:
@kbrock I added a spec (though note that I'm also trying to get the specs updated to rspec3 in a separate PR).

kbrock:
For future readers:

Using ruby `private` for a getter/setter will hide an attribute from other classes in the way desired.

Thanks @djberg96 for working this one through

```ruby
    it "honors the private option" do
      Country.data = [{:name => "US", :secret => "xxx"}, {:name => "Canada", :secret => "yyy"}]
      class Country
        private
        def secret ; self[:secret] ; end
      end

      Country.should_not respond_to(:secret)
      Country.first.send(:secret).should eq("xxx")
    end

File: lib/active_hash/base.rb [hunk 7]
```diff
@@ -218,13 +218,20 @@ def field(field_name, options = {})
...
         define_setter_method(field_name)
         define_interrogator_method(field_name)
         define_custom_find_method(field_name)
         define_custom_find_all_method(field_name)
       end
 
+      def already_defined?(method_name)
```
kbrock, line 228: @djberg96 I really like this and the changes to `instance_methods.include?` below
There is another PR out there that reports that the code sometimes overwrites methods.

Do you think it is worth it to get these changes in?

kbrock:
@djberg96 I think the already_defined? is much better than the existing solution.
think that would be nice in a pr
you're probably tired of this feature, so I can cherry-pick those into a separate pr

djberg96:
@kbrock I will leave it to you to pick what you want out of that PR. :)

PR #201, opened 2020-02-19
Support StrongParameters

rhiroe:
When an allowed `ActionController::Parameters` object was given as an argument, an empty array was returned. I wanted to get the data even with the parameters.
Even if the argument is `ActionController::Parameters` object, records will be searched correctly.

```rb
# in rails
  self.data = [
    {:id => 1, :name => "US"},
    {:id => 2, :name => "Canada"}
  ]
 end
Country.where(name: 'US').to_a
#=> [#<Country:0x0000... @attributes={:id=>1, :name=>"US"}>]

params = ActionController::Parameters.new(name: 'US')

# before
Country.where(params.permit!).to_a
#=> []

# after
Country.where(params.permit!).to_a
#=> [#<Country:0x0000... @attributes={:id=>1, :name=>"US"}>]

Country.where(params).to_a
#=> ActiveHash::ForbiddenAttributesError
```

Tell me if insufficient testing.

And `ActiveHash::Relation#where` also raise an `ArgumentError` when an unexpected object is passed as an argument.

```rb
# before
Country.where([:id, 'US'])
#=> NoMethodError (undefined method `keys' for [:name, "US"]:Array)

# after
Country.where([:id, 'US'])
#=> ArgumentError (Unsupported argument type: [:name, "US"] (Array))
```

Thanks.

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -12,6 +12,9 @@ module ActiveHash
   class FileTypeMismatchError < StandardError
   end
 
+  class ForbiddenAttributesError < StandardError
+  end
+
   class Base
 
     class_attribute :_data, :dirty, :default_attributes
--- a/lib/active_hash/relation.rb
+++ b/lib/active_hash/relation.rb
@@ -18,6 +18,12 @@ module ActiveHash
     def where(query_hash = :chain)
       return ActiveHash::Base::WhereChain.new(self) if query_hash == :chain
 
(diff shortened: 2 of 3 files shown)


kbrock (approved):
Rails throws something more similar to the previous error:

```
NoMethodError (undefined method `%' for [:name, :US]:Array)
```

But the important fact is the syntax doesn't make sense and it throws an error.

The permit part looks good

PR #202, opened 2020-02-19
add option to disable erb parsing

reedlaw:
I needed an option to disable erb parsing. My use case is the yaml is loaded at boot time but parsed at run time. There are dynamic values that need to be set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -459,6 +459,12 @@ Embedded ruby can be used in ActiveYaml using erb brackets `<% %>` and `<%= %>` 
   password: <%= ENV['USER_PASSWORD'] %>
 ```
 
+This can be disabled in an initializer:
+```ruby
+# config/initializers/active_yaml.rb
+ActiveYaml::Base.process_erb = false
+```
+
 ## ActiveJSON
 
 If you want to store your data in JSON files, just inherit from ActiveJSON and specify your path information:
--- a/lib/active_yaml/base.rb
+++ b/lib/active_yaml/base.rb
@@ -4,6 +4,10 @@ module ActiveYaml
 
   class Base < ActiveFile::Base
     extend ActiveFile::HashAndArrayFiles
+
+    cattr_accessor :process_erb, instance_accessor: false
+    @@process_erb = true
+
     class << self
       def load_file
         if (data = raw_data).is_a?(Array)
@@ -20,11 +24,15 @@ module ActiveYaml
       private
 if Psych::VERSION >= "4.0.0"
       def load_path(path)
-        YAML.unsafe_load(ERB.new(File.read(path)).result)
+        result = File.read(path)
+        result = ERB.new(result).result if process_erb
+        YAML.unsafe_load(result)
       end
 else
       def load_path(path)
-        YAML.load(ERB.new(File.read(path)).result)
+        result = File.read(path)
+        result = ERB.new(result).result if process_erb
+        YAML.load(result)
       end
 end
     end
--- a/spec/active_yaml/base_spec.rb
+++ b/spec/active_yaml/base_spec.rb
@@ -20,13 +20,26 @@ describe ActiveYaml::Base do
     Object.send :remove_const, :ArrayRow
     Object.send :remove_const, :City
     Object.send :remove_const, :State
+    Object.send :remove_const, :User
     Object.send :remove_const, :Empty
   end
 
   describe ".load_path" do
-    it 'can execute embedded ruby' do
-       expect(User.first.email).to match(/^user[0-9]*@email.com$/)
-       expect(User.first.password).to eq('secret')
+    context 'default' do
+      it 'can execute embedded ruby' do
+        expect(User.first.email).to match /^user[0-9]*@email.com$/
+        expect(User.first.password).to eq 'secret'
+      end
+    end
+
+    context 'erb disabled' do
+      before { ActiveYaml::Base.process_erb = false }
+      after  { ActiveYaml::Base.process_erb = true }
+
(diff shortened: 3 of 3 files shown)


README.md
```diff
@@ -428,6 +428,12 @@ Embedded ruby can bu used in ActiveYaml using erb brackets `<% %>` and `<%= %>`
   password: <%= ENV['USER_PASSWORD'] %>
 ```
 
+This can be disabled in an initializer:
+```ruby
+# config/initializers/active_yaml.rb
+ActiveYaml::Base.set_erb_disabled true
```
kbrock, line 434: could we keep this positive and remove the `set_`?

```suggestion
ActiveYaml::Base.process_erb = false
```

lib/active_yaml/base.rb
```diff
@@ -4,7 +4,14 @@ module ActiveYaml
 
   class Base < ActiveFile::Base
     extend ActiveFile::HashAndArrayFiles
+
+    class_attribute :erb_disabled
```
kbrock, line 8:
```suggestion
    cattr :process_erb
    @@process_erb = false
```

lib/active_yaml/base.rb
```diff
@@ -19,7 +26,9 @@ def extension
 
       private
       def load_path(path)
-        YAML.load(ERB.new(File.read(path)).result)
+        file = File.read(path)
+        result = erb_disabled ? file : ERB.new(file).result
```
kbrock, line 30: Maybe something like this?

```suggestion
        result = File.read(path)
        result = ERB.new(result).result if process_erb
```

spec/active_yaml/base_spec.rb
```diff
@@ -20,13 +20,25 @@ class Empty        < ActiveYaml::Base ; end # Empty YAML
...
-       User.first.password.should == 'secret'
+    context 'default' do
+      it 'can execute embedded ruby' do
+        User.first.email.should =~ /^user[0-9]*@email.com$/
+        User.first.password.should == 'secret'
+      end
+    end
+
+    context 'erb disabled' do
+      before { ActiveYaml::Base.set_erb_disabled true }
```
kbrock, line 36: Want to set it back otherwise we'll have sporadic test failures

```suggestion
      before { ActiveYaml::Base.process_erb = false }
      after  { ActiveYaml::Base.process_erb = true }
```

kbrock (commented):
these changes look good.

Could you let me know what you think about my suggestions and we can get this in?

Also, would you squash your commits into a single one?
thnx

lib/active_yaml/base.rb
reedlaw: Shouldn't the default be `true`?

reedlaw:
I think I addressed everything. I was not able to run the tests locally because it's been a long time since I used this. Thanks for reviewing.

lib/active_yaml/base.rb
kbrock: yes, you are right. thanks

lib/active_yaml/base.rb
```diff
@@ -4,7 +4,15 @@ module ActiveYaml
 
   class Base < ActiveFile::Base
     extend ActiveFile::HashAndArrayFiles
+
+    cattr :process_erb
```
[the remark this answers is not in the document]
kbrock: ugh (sorry, I didn't double check my syntax):

```suggestion
    cattr_accessor :process_erb, instance_accessor: false
```

lib/active_yaml/base.rb
```diff
@@ -4,7 +4,15 @@ module ActiveYaml
...
+    @@process_erb = true
+
     class << self
+      def set_erb_disabled(erb_disabled)
+        self.erb_disabled = erb_disabled
+      end
+
```
kbrock, line 15: lets go ahead and delete this

spec/active_yaml/base_spec.rb
```diff
@@ -20,13 +20,26 @@ class Empty        < ActiveYaml::Base ; end # Empty YAML
...
   end
 
   describe ".load_path" do
-    it 'can execute embedded ruby' do
-       expect(User.first.email).to match(/^user[0-9]*@email.com$/)
-       expect(User.first.password).to eq('secret')
+    context 'default' do
+      it 'can execute embedded ruby' do
+        User.first.email.should =~ /^user[0-9]*@email.com$/
+        User.first.password.should == 'secret'
```
kbrock, line 31: we are now using the expect syntax.

```suggestion
        expect(User.first.email).to match /^user[0-9]*@email.com$/
        expect(User.first.password).to eq 'secret'
```

kbrock (changes requested):
Hi @reedlaw

So sorry to do this to you.
I ran the code locally to verify that it works.

Could you make these last updates?
(maybe one of these days I'll learn how to tweak someone else's PR)

thanks
Keenan

lib/active_yaml/base.rb
```diff
@@ -24,7 +28,9 @@ def load_path(path)
       end
```
kbrock, line 28: Sigh. and then there was one...

looks like the last failure is related to Psych >= 4.0

```ruby
        result = File.read(path)
        result = ERB.new(result).result if process_erb
        YAML.unsafe_load(result)
```

Appreciate your patience especially since you are no longer using this library.

kbrock (approved):
minor convenience feature that probably will not get used by many people but good for those few who need it.

kbrock:
Thanks so much @reedlaw Esp since you no longer use this gem. Have a great week

PR #203, opened 2020-03-05
Records cannot be added to an empty ActiveJSON datastore

davidstosik:
Here's the simplest way to reproduce the error:

```rb
# test.rb

require "active_hash"

class Test < ActiveJSON::Base; end

Test.create()
```

```sh
$ echo '[]' > tests.json
$ ruby test.rb
Traceback (most recent call last):
        3: from test.rb:7:in `<main>'
        2: from /Users/david-stosik/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/active_hash-3.0.0/lib/active_hash/base.rb:174:in `create'
        1: from /Users/david-stosik/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/active_hash-3.0.0/lib/active_hash/base.rb:497:in `save'
/Users/david-stosik/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/active_hash-3.0.0/lib/active_hash/base.rb:135:in `insert': undefined method `length' for nil:NilClass (NoMethodError)
```

In some scenarios, `@records` can be `nil` when checking its `length` and appending a new record.

This PR adds a test case that reproduces the error above, and fixes it.

My first approach was to simply replace [`@records = nil`](https://github.com/zilkey/active_hash/blob/94113220327e6bb97135062e07f0d7d3429a9def/lib/active_hash/base.rb#L70) with `@records = []`, but then I figured out I could do better and improve the code base a bit, avoiding in the process duplication of code such as `@records || []`.

All specs still pass.

Please let me know what you think.

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -117,7 +117,7 @@ module ActiveHash
 
       def data=(array_of_hashes)
         mark_dirty
-        @records = nil
+        records.clear
         reset_record_index
         self._data = array_of_hashes
         if array_of_hashes
@@ -135,13 +135,12 @@ module ActiveHash
       end
 
       def insert(record)
-        @records ||= []
         record[:id] ||= next_id
         validate_unique_id(record) if dirty
         mark_dirty
 
-        add_to_record_index({ record.id.to_s => @records.length })
-        @records << record
+        add_to_record_index({ record.id.to_s => records.length })
+        records << record
       end
 
       def next_id
@@ -152,6 +151,12 @@ module ActiveHash
           max_record.id.succ
         end
       end
+
+      def records
+        @records ||= []
+      end
+
+      private :records
 
       def record_index
         @record_index ||= {}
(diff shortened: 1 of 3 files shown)


kbrock (commented):
this looks nice.

consolidates the delete_all details.
not sure that `records` buys us that much but good enough

could you rebase this to fixup your spec and then we can get this merged

davidstosik:
Sorry for the delay, this is rebased now. 👌

davidstosik:
This still seems to be a problem, so I rebased my branch.

kbrock:
Darn. this is still outstanding. After we merge #268 I'd like to get this in.
Please ping me if I haven't gotten this in by 5/15/2023

kbrock:
I thought we had a whole discussion here.

I view active_hash as a static data store. And if you want to treat it as a writable database, there are other database projects that may be better suited (like sqlite).

sqlite does support `sqlite3 dbfile.sqlite ".dump"` which will output the results as text. This may meet your needs.

flavorjones:
@kbrock I agree. We did discuss this a bit at #234 and I think we should decline to address writing behaviors.

PR #207, opened 2020-05-15
Implement ActiveRecord::RecordNotFound interface

ChrisBr:
We should include `model`, `id` and `primary_key`
attributes too to have the same interface as `ActiveRecord::RecordNotFound`.

This would make it possible to do something like this

```ruby
rescue_from ActiveRecord::RecordNotFound, ActiveHash::RecordNotFound, :render_not_found

def not_found
  render_error :record_not_found, message: "#{exception.model} not found.", status: :not_found
end
```

### DIFF
--- a/spec/active_hash/base_spec.rb
+++ b/spec/active_hash/base_spec.rb
@@ -598,9 +598,15 @@ describe ActiveHash, "Base" do
       end
 
       it "raises ActiveHash::RecordNotFound when id not found" do
-        expect do
-          Country.find(0)
-        end.to raise_error(ActiveHash::RecordNotFound, /Couldn't find Country with ID=0/)
+        expect { 
+          Country.find(0) 
+        }.to raise_error(an_instance_of(ActiveHash::RecordNotFound)
+          .and having_attributes(
+            message: "Couldn't find Country with ID=0",
+            primary_key: 'id',
+            id: 0
+          )
+        )
       end
     end
 
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -1,6 +1,14 @@
 module ActiveHash
-
   class RecordNotFound < StandardError
+    attr_reader :model, :primary_key, :id
+
+    def initialize(message = nil, model = nil, primary_key = nil, id = nil)
+      @primary_key = primary_key
+      @model = model
+      @id = id
+
+      super(message)
+    end
   end
 
   class ReservedFieldError < StandardError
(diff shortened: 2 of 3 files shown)


spec/active_hash/base_spec.rb
```diff
@@ -594,9 +594,15 @@ class Region < ActiveHash::Base
       end
 
       it "raises ActiveHash::RecordNotFound when id not found" do
-        proc do
+        begin
```
ChrisBr, line 597: Seems like rspec 2.2 does not support `have_attributes` for `raise_error` so I did this rather ugly thing here. Open for better solutions though :/

```ruby
RSpec.describe "composing matchers" do
  it "raises StandardError" do
    expect { raise StandardError, "my message" }.
      to raise_error(an_instance_of(StandardError).and having_attributes({"message" => "my message"}))
  end
end
```

https://relishapp.com/rspec/rspec-expectations/v/2-2/docs/matchers/expect-error
https://relishapp.com/rspec/rspec-expectations/v/3-9/docs/built-in-matchers/raise-error-matcher#set-expectations-on-error-object-with-chained-matchers

djberg96:
@ChrisBr I would vote to update the specs to rspec 3. There's a conversion tool called transpec I've used in the past that worked well: https://github.com/yujinakayama/transpec

ChrisBr:
> @ChrisBr I would vote to update the specs to rspec 3. There's a conversion tool called transpec I've used in the past that worked well: https://github.com/yujinakayama/transpec

I think this is a bit outside scope of this PR? 🤔

djberg96:
@ChrisBr Oh, definitely, I was just suggesting that happen first in a separate PR, then go from there.

I'm actually trying to do it now but am running into some sort of bundler conflict....

kbrock:
@ChrisBr Can you either update, or if you don't have the time, I may be able to rebase this for you and merge.

ChrisBr:
> Can you either update, or if you don't have the time, I may be able to rebase this for you and merge.

Rebased @kbrock 👍

ChrisBr:
@kbrock I think we need to update the RSpec version in the gemfiles directory too.

https://github.com/zilkey/active_hash/pull/213

kbrock:
rerunning the builds and hoping that we will be all set on that one soon

ChrisBr:
> rerunning the builds and hoping that we will be all set on that one soon

Seemed to work 👍 Thanks for a lot of merges the last couple of days 👏👏👏

kbrock:
thanks for helping us move forward.
You've been giving us some good stuff

PR #209, opened 2020-08-03
Switch to rspec3 (and update the Gemfile)

djberg96:
This PR switches the rspec tests from version 2 to version 3, i.e. the current version at the time of this PR.

I also updated the Gemfile. In addition to the rspec version update, I removed the json check for a Ruby version less than 2 because the gemspec already mandates Ruby 2.4 or later. Also, I removed the version dependency for rake, as it really shouldn't matter.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ gemspec
 
 gem 'rspec', '~> 3.9'
 gem 'wwtd'
-gem 'rake', '~> 10.0'
+gem 'rake'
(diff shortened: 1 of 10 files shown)


ChrisBr:
Should we update the Gemfile's in https://github.com/zilkey/active_hash/tree/master/gemfiles too? cc @kbrock

ChrisBr:
Related to https://github.com/zilkey/active_hash/pull/213

PR #212, opened 2020-09-10
Add model class and id to RecordNotFound error

mizukami234:
Added model class and id attributes to ActiveHash::RecordNotFound error for some situations that need more specific handling, like [ActiveRecord::RecordNotFound](https://github.com/rails/rails/blob/070d4afacd3e9721b7e3a4634e4d026b5fa2c32c/activerecord/lib/active_record/errors.rb#L60).

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -1,6 +1,13 @@
 module ActiveHash
 
   class RecordNotFound < StandardError
+    attr_reader :model, :id
+
+    def initialize(message = nil, model = nil, id = nil)
+      @model = model
+      @id = id
+      super(message)
+    end
   end
 
   class ReservedFieldError < StandardError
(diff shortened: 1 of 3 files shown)


lib/active_hash/base.rb
```diff
@@ -1,6 +1,13 @@
 module ActiveHash
 
   class RecordNotFound < StandardError
+    attr_reader :model, :id
+
+    def initialize(message = nil, model = nil, id = nil)
+      @model = model
+      @id = id
+      super(message)
+    end
```
kbrock, line 10: is there a reason to not copy the complete `ActiveRecord::RecordNotFound` interface?

```ruby
  class RecordNotFound < ActiveRecordError
    attr_reader :model, :primary_key, :id

    def initialize(message = nil, model = nil, primary_key = nil, id = nil)
      @primary_key = primary_key
      @model = model
      @id = id

      super(message)
    end
  end
```

 seems to make sense to  mimic rails completely.

kbrock (commented):
LGTM

just that one question/change

kbrock:
Looks like this is the same as #207 

closing as a duplicate

PR #218, opened 2020-11-12
Added primary_key support for has_one same as has_many

yujideveloper:
I added `primary_key` support for `has_one` association that is same as `has_many` association. ref. #71
And also tested `foreign_key` support explicitly.

### DIFF
--- a/lib/associations/associations.rb
+++ b/lib/associations/associations.rb
@@ -116,7 +116,8 @@ module ActiveHash
         define_method(association_id) do
           options = {
             :class_name => association_id.to_s.classify,
-            :foreign_key => self.class.to_s.foreign_key
+            :foreign_key => self.class.to_s.foreign_key,
+            :primary_key => self.class.primary_key
           }.merge(options)
 
           scope = options[:class_name].constantize
@@ -124,7 +125,7 @@ module ActiveHash
           if scope.respond_to?(:scoped) && options[:conditions]
             scope = scope.scoped(:conditions => options[:conditions])
           end
-          scope.send("find_by_#{options[:foreign_key]}", id)
+          scope.send("find_by_#{options[:foreign_key]}", send(options[:primary_key]))
         end
       end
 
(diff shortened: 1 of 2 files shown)


yujideveloper:
@kbrock 
What do you think this pr?

yujideveloper:
@kbrock 
Cloud you review this?
I think this is useful.

lib/associations/associations.rb
```diff
@@ -116,15 +116,16 @@ def has_one(association_id, options = {})
...
+            :primary_key => self.class.primary_key
           }.merge(options)
 
           scope = options[:class_name].constantize
 
           if scope.respond_to?(:scoped) && options[:conditions]
             scope = scope.scoped(:conditions => options[:conditions])
           end
-          scope.send("find_by_#{options[:foreign_key]}", id)
+          scope.send("find_by_#{options[:foreign_key]}", send(options[:primary_key]))
```
kbrock, line 128: Specifying the primary key looks strange, but that is part of the rails api.

(expanding this PR) Is there a reason why we wouldn't use `find_by()`?

```suggestion
          scope.find_by(options[:foreign_key] => send(options[:primary_key]))
```

kbrock: ooh, I think `read_attribute(options[:primary_key])` would be preferred over the `send(options[:primary_key])`

yujideveloper: @kbrock 
I think so too, but I prioritized making the code the same as `belongs_to` and `has_many` to avoid unexpected behavior in this PR.
https://github.com/active-hash/active_hash/blob/5d341e57036b67985734de5fd932da1c6d6f8f72/lib/associations/associations.rb#L102 https://github.com/active-hash/active_hash/blob/5d341e57036b67985734de5fd932da1c6d6f8f72/lib/associations/associations.rb#L142

I will try to refactor using `find_by` and `read_attribute` to `has_one` / `has_many` / `belongs_to` in the next other PR.
What do you think of this plan?

yujideveloper: I lightly tried using `read_attribute` / `find_by`, but there are where it does not work.
ActiveHash is stored attributes to plain `Hash`, and `read_attribute` / `find_by` are not converting key. So, `Symbol` key and `String` key are not considered to be the same.

kbrock:
passes test locally.

This is a standard rails method, following the same pattern

PR #220, opened 2021-01-03
docs: Fix the example to work

tbotaq:
Hi.

The following code of `ActiveFile::Base.full_path` shows that the first dot of file extension is not needed when a programmer overrides `ActiveFile::Base.extension`.
https://github.com/zilkey/active_hash/blob/1a07f3ea7b4920f35843010cc515be8c8695e612/lib/active_file/base.rb#L36

So I'm removing the first dot from the example code to go with the implementation of `ActiveFile::Base.full_path`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -492,7 +492,7 @@ class Country < ActiveFile::Base
 
   class << self
     def extension
-      ".super_secret"
+      "super_secret"
     end
 
     def load_file


README.md
```diff
@@ -492,7 +492,7 @@ class Country < ActiveFile::Base
 
   class << self
     def extension
-      ".super_secret"
+      "super_secret"
```
tbotaq, line 495: Adding a dot at the first part of `extension` leads our users to seeing some kind of the following error.
```
No such file or directory @ rb_sysopen - /foo/bar/some_file..super_secret (Errno::ENOENT)
```

So removing the dot from the example above may be a little kinder for our users, I think.

PR #228, opened 2021-04-01
Add an expectation to a test

tbotaq:
The original test works well enough as a test for [the commit](https://github.com/zilkey/active_hash/commit/209834d39f688597f0f85cebe69d30e8c6ae48c5) .

But it might be a little clearer for its readers if the test has its own expectation.
So I added an expectation that shows us what the commit did, tweaking its description as well.

### DIFF
--- a/spec/associations/active_record_extensions_spec.rb
+++ b/spec/associations/active_record_extensions_spec.rb
@@ -177,7 +177,7 @@ unless SKIP_ACTIVE_RECORD
           expect(school.city).to eq(city)
         end
 
-        it "returns nil when the belongs_to association class can't be autoloaded" do
+        it "doesn't raise any exception when the belongs_to association class can't be autoloaded" do
           # Simulate autoloader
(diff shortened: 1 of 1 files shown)


spec/associations/active_record_extensions_spec.rb
```diff
@@ -177,10 +177,10 @@ class Book < ActiveRecord::Base
           expect(school.city).to eq(city)
         end
 
-        it "returns nil when the belongs_to association class can't be autoloaded" do
+        it "doesn't raise any exception when the belongs_to association class can't be autoloaded" do
```
tbotaq, line 180: According to [the commit](https://github.com/zilkey/active_hash/commit/209834d39f688597f0f85cebe69d30e8c6ae48c5) that added this test,
- It doesn't return `nil` in this case
- The purpose of the commit was to rescue a LoadError, so this test might have to check if it doesn't raise any exception, not its return value.

kbrock:
thank you

tbotaq:
Thank you, too !

PR #229, opened 2021-06-05
Add thread safety to ActiveFile

### LINKED ISSUE #159: Thread Safety
Wrote simplistic example to evaluate below: 

```ruby
require 'active_hash'
class Item < ActiveYaml::Base
end
@thr = []
def f(i) 
    @thr << Thread.new { puts 'Thread ' + i.to_s + ': ' + Item.all.length.to_s }
end
for i in 0..5
    f i
end
@thr.each(&:join)
``` 

Most of the time output is `Thread 5: 4, Thread 0: 4, Thread 3: 4, Thread 2: 4, Thread 1: 4, Thread 4: 4` with different thread ordering, but sometimes I see `Thread 3: 0, Thread 2: 0, Thread 4: 0, Thread 1: 0
, Thread 0: 4, Thread 5: 4` 

I have `items.yml` file in the root with 4 items `- val: 1`. My `Gemfile.lock` shows the `active_hash (1.5.3)` which I suppose is the latest.

I would expect for library to guarantee thread safety.

dmitriy-kiriyenko:
ActiveFile is not thread-safe, which results in the following: if the very first request to ActiveFile goes simultaneously from even two threads, the first one is going to return as if the data was empty. The second request will immediately fix that, but that unlucky thread will receive a bad value. It may store it in the database or memoize it, which will result in further problem escalation.

We can't just synchronise it, because we check the variable very far from reload. While we must synchronise, it's not enough. We have to flag it as 'done' only after it's loaded.

Another problem is that we touch `all` in process of reloading, which will check `data_loaded` flag and try to reload again. To counter that, I come out with a method that won't trigger reload, to use it in methods that work during reload.

Resolves #159.

### DIFF
--- a/lib/active_file/base.rb
+++ b/lib/active_file/base.rb
@@ -2,6 +2,7 @@ module ActiveFile
 
   class Base < ActiveHash::Base
     extend ActiveFile::MultipleFiles
+    @@instance_lock = Mutex.new
 
     class_attribute :filename, :root_path, :data_loaded, instance_reader: false, instance_writer: false
 
@@ -13,10 +14,12 @@ module ActiveFile
       end
 
       def reload(force = false)
-        return if !self.dirty && !force && self.data_loaded
-        self.data_loaded = true
-        self.data = load_file
-        mark_clean
+        @@instance_lock.synchronize do
+          return if !self.dirty && !force && self.data_loaded
+          self.data = load_file
+          mark_clean
+          self.data_loaded = true
+        end
       end
 
       def set_filename(name)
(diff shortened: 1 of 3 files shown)


dmitriy-kiriyenko:
@zilkey any chance?

yujideveloper:
@adampal @kbrock 
I believe we should merge this pr.
Could you review this?

adampal:
Thanks @dmitriy-kiriyenko I tested locally and confirmed this definitely fixes the issue @drmnick saw in #159

adampal (approved):

PR #230, opened 2021-06-17
support i18n

ryu-sato:
Hi, I am hoping that ActiveHash will support i18n.

Since the reason why #170 does not merge for a long time is that it needs tests, I created a new PR that adds tests based on #170.

- [x] `bundle exec rspec` passed


kbrock:
I think a test would be to add a localization file in specs and verify that the translated name comes through.

ryu-sato:
@kbrock Thank you for your feedback. I have addressed your feedback.

PR #231, opened 2021-07-06
Speed up automatic id definition

braktar:
In our use case, we have a large amount of data with no id. It generates a performance issue due to the comparison of all the ids to find the maximum.

To fix it, we store the current maximum, and only perform the comparison to this value during the insertion.

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -136,17 +136,21 @@ module ActiveHash
 
       def insert(record)
         @records ||= []
-        @max_id ||= 0
-        if record[:id] && record[:id].is_a?(Numeric)
-          @max_id = [@max_id, record[:id]].max
-        else
-          record[:id] ||= (@max_id = @max_id.succ)
-        end
+        set_id(record)
         validate_unique_id(record) if dirty
         mark_dirty
 
         add_to_record_index({ record.id.to_s => @records.length })
         @records << record
+      end
+
+      def set_id(record)
+        # sets record[:id] to @max_id+1 if it doesn't exist
+        if record[:id] && record[:id].is_a?(Numeric)
+          @max_id = [@max_id || 0, record[:id].ceil].max
+        else
+          record[:id] ||= (@max_id = @max_id.to_i.succ)
+        end
       end
 
       def record_index
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -136,21 +136,17 @@ module ActiveHash
 
       def insert(record)
         @records ||= []
-        record[:id] ||= next_id
+        @max_id ||= 0
+        if record[:id] && record[:id].is_a?(Numeric)
+          @max_id = [@max_id, record[:id]].max
+        else
+          record[:id] ||= (@max_id = @max_id.succ)
+        end
         validate_unique_id(record) if dirty
         mark_dirty
 
         add_to_record_index({ record.id.to_s => @records.length })
         @records << record
-      end
-
-      def next_id
-        max_record = all.max { |a, b| a.id <=> b.id }
-        if max_record.nil?
-          1
-        elsif max_record.id.is_a?(Numeric)
-          max_record.id.succ
-        end
       end
 
       def record_index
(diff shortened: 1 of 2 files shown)


lib/active_hash/base.rb
```diff
@@ -136,23 +136,19 @@ def exists?(record)
...
       def insert(record)
         @records ||= []
-        record[:id] ||= next_id
+        @max_id ||= 0
+        if record[:id] && record[:id].is_a?(Numeric)
+          @max_id = [@max_id, record[:id]].max
+        else
+          record[:id] ||= (@max_id = @max_id.succ)
+        end
         validate_unique_id(record) if dirty
```
senhalil, line 145: It is possible to create an object with a negative id. In this case the `max_id` doesn't get updated correctly by the suggested edit. The previous behaviour would correctly identify the maximum id in this case.

There was also an issue with the previous code that `.succ` exists only for `Integer` and not for `Numeric` so it would raise a `NoMethodError Exception: undefined method 'succ' for 1.0:Float` if an object with a `Float` id happens to have the maximum `id`. This issue persists in the suggested edit.

I think it might be better to keep the id logic in a separate function to not clutter the insert logic. So I suggest the following (untested) code:

```ruby
      def insert(record)
        @records ||= []
        set_id(record)
        validate_unique_id(record) if dirty
        ...
      end

      begin set_id(record)
        # sets record[:id] to @max_id+1 if it doesn't exist
        if record[:id] && record[:id].is_a?(Numeric)
          @max_id = [@max_id || -Float::INFINITY, record[:id].ceil].max
        else
          record[:id] ||= (@max_id = @max_id.to_i.succ)
        end
      end
```
@braktar let me know if you like the suggested edit. Either I can include it in my PR or commit to your branch.

braktar: I just prefer to be consistent with the current tests and initialize @max_id to 0

senhalil: In that case can you add a test that creates an object with negative ID and then another one without id, to check the id of the last one to be 1.

You still take care of the Float id problem which can be handled with `.ceil` .

senhalil: Oh I just saw the new commit, you already did the test :facepalm:

lib/active_hash/base.rb
```diff
@@ -136,19 +136,23 @@ def exists?(record)
...
         @records << record
       end
 
+      def set_id(record)
+        # sets record[:id] to @max_id+1 if it doesn't exist
+        if record[:id] && record[:id].is_a?(Numeric)
+          @max_id = [@max_id || 0, record[:id].ceil].max
+        else
+          record[:id] ||= (@max_id = @max_id.to_i.succ)
+        end
```
senhalil: I didn't initialise `@max_id` to handle the negative values but if you are okay with ignoring the negative IDs, then you can be more expressive (similar to your original PR) -- no need for hiding `nil.to_i = 0` in `@max_id.to_i.succ` and  `[@max_id || 0, ...].max`.

```ruby
      def set_id(record)
        # sets record[:id] to @max_id+1 if it doesn't exist
        @max_id ||= 0
        if record[:id] && record[:id].is_a?(Numeric)
          @max_id = [@max_id, record[:id].ceil].max
        else
          record[:id] ||= (@max_id = @max_id.succ)
        end
      end
```

senhalil (approved):

senhalil:
When the warnings are enabled, `set_id` spits a message because of `@max_id`. 

I don't know which gem has changed the defaults but after a `bundle update` `rake test` gives the following error on our optimizer-api repository. 
Screenshot from 2021-07-20 16-54-35

I suggest the following modification 
```ruby
      def set_id(record)
        # sets record[:id] to @max_id+1 if it doesn't exist
        @max_id ||= 0
        if record[:id] && record[:id].is_a?(Numeric)
          @max_id = [@max_id, record[:id].ceil].max
        else
          record[:id] ||= (@max_id = @max_id.succ)
        end
      end
```

kbrock:
We should not be generating ids or values in Active Hash, so we're going to avoid optimizing and/or encouraging that behavior.

I think this behavior does meets your needs well. So you may want to add these 2 methods to a module and prepend in your class.

Issue #225, opened 2021-02-08
Release master for Ruby 3 compatibility


nsommer:
Would you consider bumping the version and doing a release of master? ActiveHash currently does not run with Ruby 3, but it was already fixed on master (https://github.com/zilkey/active_hash/commit/f6448312c748f7e0f2846ed0d377074c974d3f7d).

nsommer:
@kbrock @syguer any chance you would release soonish? 🙂

kbrock:
Hi @syguer -- are you the one to push these changes live?
Anything other people need to do to help with this effort?

aknicol:
Have these issues for Ruby 3 been resolved? I'm still receiving `Undefined method 'arity'` on my model definitions.

aknicol:
As I mentioned I'm still having issues, my code follows the example in the docs almost exactly, I have: 

```
class Person < ActiveRecord::Base

  extend ActiveHash::Associations::ActiveRecordExtensions
  belongs_to_active_hash :country

```
 
on the next line, 

`  belongs_to :surname, optional: true
`
I get the error **undefined method `arity' for {:optional=>true}:Hash**

cath-oneill:
Finally upgrading to Ruby 3, and running into the same issues.  It does appeared to be fixed for my use-case both in master ([f644831](https://github.com/zilkey/active_hash/commit/f6448312c748f7e0f2846ed0d377074c974d3f7d)) as @nsommer pointed out, and @gazayas' offered another possible fix in his PR.  Anything that can be done to help get the work that's already been done released?

Sixeight:
@zilkey Could you release a new version? Current version has some problem with Ruby 3.x.
Or, How about an idea create new organization for this library to maintain the software by community.
Thanks for the great library!

Sixeight:
https://github.com/active-hash had been created. but the new version has not been released. does anyone has manageable permission?

aknicol:
Has there been any progress on this?

yujideveloper:
v3.1.1 has been released that supports Ruby 3.x.
https://rubygems.org/gems/active_hash/versions/3.1.1

neocoin:
Please release version v3.1.2.  

I've made some monkey patches to make active_hash work in Ruby 3.2 and Rails 7. 

For example, I must need these commits.

00f98114f9b3c22010a47f3f7b35b7aed1a33b5d
bb111597fc2214f02670fb5d2d35f374a5f98466

[...]

PR #240, opened 2021-11-12
Use double splat operator in belongs_to call to super

gazayas:
I ran across this in a project I'm working on and found out other people were having the same problem: Issue #225.

`super` was calling the following method in ActiveRecord:
```ruby
def belongs_to(name, scope = nil, **options)
  reflection = Builder::BelongsTo.build(self, name, scope, options)
  Reflection.add_reflection self, name, reflection
end
```

`options` was getting substituted into `scope` which was causing the bug.
There might be a better fix for this (Using [ruby2_keywords](https://github.com/ruby/ruby2_keywords)?) so I can look over the code just in case, but this fixes the issue for me.

### DIFF
--- a/lib/associations/associations.rb
+++ b/lib/associations/associations.rb
@@ -16,7 +16,7 @@ module ActiveHash
         if klass && klass < ActiveHash::Base
           belongs_to_active_hash(name, options)
         else
-          super
+          super(name, **options)
         end
       end
 


cath-oneill:
This was fixed in ([f644831](https://github.com/zilkey/active_hash/commit/f6448312c748f7e0f2846ed0d377074c974d3f7d)), but hasn't been included in a versioned release yet.

PR #246, opened 2022-04-06
Gem versions

kbrock:
## Overview

I made the test run locally on my machine.

## Why fix travis.yml

I don't have the privs to migrate us from travis to github actions.
But I can fix some test issues.

Also, this will hopefully remove work for when we want to add the action.

## What changed

Each one has a commit message with more details:

- updated for changes to ruby 3.1.1 and psyche issues
- changed the `should` to `expect` tests
- removed `autorun` as it has been deprecated for a while
- only run tests with supported rails and ruby versions. (please holler if you want to support EOL ruby/rails)

## What was out of scope

While I did add a rails 7.0 gemspec, fixing the suite was out of scope.
Once that runs, we can uncomment the line in the travis.yml


kbrock:
@syguer I had merged a few PRs and wanted to make sure that the suite was still green.
When I downloaded to run it, things were in a little off.
Hope you like these changes.

I can merge if you are too busy, but would prefer to get your ok.

kbrock:
update:
- still support deprecated versions of rails and ruby. But don't run so many combinations on EOL versions

lib/active_yaml/base.rb
```diff
@@ -18,9 +18,15 @@ def extension
       end
 
       private
+if RUBY_VERSION > "3.1.0"
```
yujideveloper, line 21: @kbrock 
I guess this change is to support Psych v4, but Psych v4 is also available in <Ruby 3.1. So, I think we should use `Psych::VERSION` instead of `RUBY_VERSION`.

kbrock:
update:
- changed version checking to be Psych based (and not ruby based)

kbrock:
@yujideveloper do you want me to pull out the rspec changes (`autorun` and `expect`) for an easy merge?

Do you also consider the `YAML` change an easy merge?

While we can not run these on travis, it would be nice to upgrade the `.travis.yml` to at least run from the local command line.

lib/active_yaml/base.rb
```diff
@@ -18,9 +18,15 @@ def extension
       end
 
       private
+if Psych::VERSION >= "4.0.0"
+      def load_path(path)
+        YAML.load(ERB.new(File.read(path)).result, aliases: true, permitted_classes: [Time, Date, Symbol])
```
yujideveloper, line 23: We may need to consider using `unsafe_load` for backwards compatibility.
`YAML.unsafe_load` in Psych 4.0 is same as `YAML.load` in Psych < 4.0.

yujideveloper:
@kbrock 
> do you want me to pull out the rspec changes (autorun and expect) for an easy merge?

I think it's better to pull out, but I think it's okay as it is.

> Do you also consider the YAML change an easy merge?

I think it is easy merge, but I have one concern in https://github.com/active-hash/active_hash/pull/246#discussion_r853352555.

lib/active_yaml/base.rb
kbrock: ok, so just use `unsafe_load` and remove the need for the `permitted_classes` and gook

kbrock:
update:

- extracted changes to another PR
- changed `YAML.load` to `YAML.unsafe_load`
- removed those commits from here

adampal:
@kbrock now that we've cut over to GitHub Actions, is this PR still required?

adampal (approved):

Issue #247, opened 2022-04-14
Scopes are not chainable


andreynering:
Hey there!

Thanks for this project! We use it extensively at JobScore and love it.

I'd like to request something, though. Scopes work in isolation, but you can't use more than one together, which works in ActiveRecord and is extremely convenient. Could this possibly be implemented?

```ruby
class Vehicle
  scope :car, -> { where(type: 'car' }
  scope :motorized, -> { where(motorized: true}
end
```

```ruby
> Vehicle.car.motorized
NoMethodError: undefined method `motorized' for #<ActiveHash::Relation:0x000000010732d9c8>
```

andreynering:
I just gave it a try at https://github.com/active-hash/active_hash/pull/248

PR #248, opened 2022-04-14
ActiveHash::Relation: Allow chaining multiples scopes

andreynering:
Closes https://github.com/active-hash/active_hash/issues/247

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -22,7 +22,7 @@ module ActiveHash
 
   class Base
 
-    class_attribute :_data, :dirty, :default_attributes
+    class_attribute :_data, :dirty, :default_attributes, :scopes
 
     class WhereChain
       def initialize(scope)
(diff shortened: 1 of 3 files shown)


kbrock (commented):
passes all specs.

I like the idea of adding scopes to the chain.
Not so sure about the `method_missing` approach. But to be honest, that is basically the way scopes are implemented, so I don't see any other way.

Anyone else have an opinion on the `method_missing` and another approach?
If no one does, then I propose merging

andreynering:
@kbrock Yeah, I used `method_missing` because I was out of other ideas on how to implement that.

If anyone has a suggestion, let me know. But this was probably the way to do it.

adampal:
Nice work @andreynering 
I agree that scopes should be chainable so the interface is more consistent with ActiveRecord.

I don't have a better suggestion than what you've done with method missing.  I'm happy to merge this.

andreynering:
Hi @adampal,

Thanks for accepting my contribution! ❤️ 

I'd just like to remind you that you should probably push a new release soon so users can actually use this and other improvements made since the last release, which was a long time ago, about 2 and half years ago.

adampal:
Hey @andreynering no worries, I'll publish a new release shortly. Just wanting to finalise a couple of other pieces first.
I think we'll have the new release ready to go within a week or so.

kbrock:
thanks for the backup review

PR #250, opened 2022-04-21
Source tweaks

kbrock:
pulled out of #246 

These are base changes that help various versions of the gems run better with active_hash


kbrock:
If there is something holding this up, please let me know and I can fix.

I would really like this and the travis.yml PRs merged. So when merging other's code, I at least have a green test suite that I can run locally.

Do we have a rule against merging our own code?

PR #258, opened 2022-08-16
Fix rspec config when SKIP_ACTIVE_RECORD enabled

yujideveloper:
I fixed rspec config when `SKIP_ACTIVE_RECORD` enabled.
It broke at #256


kbrock (approved):
This looks good

Could you update the description to explain why this fixes?
The tests looked green before and after this change

yujideveloper:
I updated PR's description.
And I rebased my branch.

PR #259, opened 2022-08-16
Fix deprecation warnings

yujideveloper:
I fixed two deprecation warnings below. 

* `DEPRECATION WARNING: Rails 7.0 has deprecated Enumerable.sum in favor of Ruby's native implementation available since 2.4. Sum of non-numeric elements requires an initial argument.
`
* `DEPRECATION WARNING: Time#to_s(:number) is deprecated. Please use Time#to_fs(:number) instead.`

### DIFF
--- a/lib/active_file/hash_and_array_files.rb
+++ b/lib/active_file/hash_and_array_files.rb
@@ -13,7 +13,7 @@ module ActiveFile
       loaded_files = full_paths.collect { |path| load_path(path) }
 
       if loaded_files.all?{ |file_data| file_data.is_a?(Array) }
-        loaded_files.sum
+        loaded_files.sum([])
       elsif loaded_files.all?{ |file_data| file_data.is_a?(Hash) }
         loaded_files.inject({}) { |hash, file_data| hash.merge(file_data) }
       else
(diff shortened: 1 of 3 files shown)


lib/active_file/hash_and_array_files.rb
```diff
@@ -13,7 +13,7 @@ def data_from_multiple_files
       loaded_files = full_paths.collect { |path| load_path(path) }
 
       if loaded_files.all?{ |file_data| file_data.is_a?(Array) }
-        loaded_files.sum
```
kbrock, line 16: is this something we can fix in the definition of `sum`?

Or is this just a place where the method signature has been messed up in upstream rails?

kbrock (approved):

kbrock:
@yujideveloper do you know what the business process is around the review/merge?

Am I allowed to review and merge your changes?
Or do we need 2 different people on that?
With this project, I worry that requiring 2 different reviewers would mean nothing gets merged

yujideveloper:
@kbrock 

Sorry, late reply.

> do you know what the business process is around the review/merge?

I don't know it in this project.

> Am I allowed to review and merge your changes?
> Or do we need 2 different people on that?
> With this project, I worry that requiring 2 different reviewers would mean nothing gets merged

I think it's better to have two different reviewers. But as you say, I think nothing will be merged in this project now.

Issue #257, opened 2022-07-26
ActiveHash scope affecting self


mogya:
Define scope on ActiveHash class and calling it, seems polluting original object.

# reproduce
```
class Country < ActiveHash::Base
  self.data = [
    {:id => 1, :name => "US"},
    {:id => 2, :name => "Japan"}
  ]
  scope :east_asia, -> do
    where(name: "Japan")
  end
end

irb(main):054:0> c = Country.all
=> 
#<ActiveHash::Relation:...
irb(main):055:0> c.count
=> 2
irb(main):056:0> c.east_asia
=> 
#<ActiveHash::Relation:...
irb(main):057:0> c.count
=> 1
```

# expected behavior

c.count should returns 2 because ActiveRecord scope returns new relation and do not affect to self.

# environment

```
irb(main):058:0> ActiveHash::gem::VERSION
=> "3.1.1"
irb(main):063:0> Rails::VERSION::STRING
=> "7.0.3.1"
```

pfeiffer:
The same happens with simply using `where` on a scope. A failing test:

```ruby
countries = Country.where(language: "English")

expect(countries.size).to eq 2
expect(countries.where(name: "US").size).to eq 1
expect(countries.size).to eq 2 🤯 # fails as the previous line mutated the scope!

pfeiffer:
Fixed in #265

kbrock:
Going with #268

PR #261, opened 2022-08-25
Fix order bug

### LINKED ISSUE #193: The `order` method breaks finders

Yuma-Nagaoka:
# Overview
fix #193

`ActiveHash::Relation#find` can't return a correct record after executing `#order`.
The reason is that the `#order_by_args!` method destructively reorders the 'relation' itself, and `#find` is based on the 'relation's index' which has the wrong reordered index.

# How?
Change `#order_by_args!` method to be non-destructive.

### DIFF
--- a/lib/active_hash/relation.rb
+++ b/lib/active_hash/relation.rb
@@ -96,9 +96,7 @@ module ActiveHash
       processed_args = preprocess_order_args(options)
       candidates = relation.dup
 
-      order_by_args!(candidates, processed_args)
-
-      candidates
+      order_by_args(candidates, processed_args)
     end
 
     def to_ary
@@ -179,7 +177,9 @@ module ActiveHash
       ary.map! { |e| e.split(/\W+/) }.reverse!
     end
 
-    def order_by_args!(candidates, args)
+    def order_by_args(candidates, args)
+      ordered_candidates = []
+
       args.each do |arg|
         field, dir = if arg.is_a?(Hash)
                        arg.to_a.flatten.map(&:to_sym)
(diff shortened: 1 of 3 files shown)


pfeiffer:
This is much needed - mutating the original object is really unexpected. See also #257 and #265.

kbrock:
@Yuma-Nagaoka Please add a test that sorts by `order([])`. Think this will return an empty array.

@flavorjones I feel like the other implementation is closer to active record, where it does a `dup` and all methods seem to modify the value in place (i.e.: are bang methods). But you are right. this is simpler.

This looks good. Just want that one test to make sure it is not breaking things if args is empty.

Also, not sure how the original code worked with 2 arguments for sorting `order(:id, :name)`. In other implementations I always have had to map the args and compare those arrays (worrying about nils sneaking in). This existing solution looks like it is too simple.

kbrock:
Yea, ok. so my thoughts were right the `order(nil)` and `order({})` return `[]`
- spec/active_hash/base_spec.rb:940
- spec/active_hash/base_spec.rb:944

Also, it looks like it screwed up the original order as well. These 2 tests are returning `US` instead of `Canada`:

- spec/active_hash/base_spec.rb:963
- spec/active_hash/base_spec.rb:984

kbrock:

PR #262, opened 2022-09-07
add ActiveRecord like exists? behavior.

ise-tang:
Currently `ActiveHash::Base` has `exists?` method but its behavior is different with ActiveRecord's one.  
It seems like to be used in `save` method only but it is a public method.
So I added ActiveRecord like behaviors and keep current behavior.

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -133,11 +133,11 @@ module ActiveHash
         elsif args === false
           false
         elsif args.nil?
-          all.count.positive?
+          all.present?
         elsif args.class == Hash
-          all.where(args).count.positive?
+          all.where(args).present?
         else
-          all.where(id: args.to_i).count.positive?
+          all.where(id: args.to_i).present?
         end
       end
 
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -127,9 +127,17 @@ module ActiveHash
         end
       end
 
-      def exists?(record)
-        if record.id.present?
-          record_index[record.id.to_s].present?
+      def exists?(args = nil)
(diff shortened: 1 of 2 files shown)


ise-tang:
@kbrock How about this? If something lacks for your decision, please tell me.

lib/active_hash/base.rb
```diff
@@ -127,9 +127,17 @@ def data=(array_of_hashes)
...
-          record_index[record.id.to_s].present?
+      def exists?(args = nil)
+        if args.respond_to?(:id)
+          record_index[args.id.to_s].present?
+        elsif args === false
+          false
+        elsif args.nil?
+          all.count.positive?
+        elsif args.class == Hash
+          all.where(args).count.positive?
```
kbrock, line 138: is there a reason to pick `all.where(args).count.positive?` vs `all.where(args).present?`

ise-tang: The reason why I picked it is I misunderstood `where(args)` like AR object. 
I tested both by the follow script, then I know `present?` is faster than `count.positive?` 
So I haver replaced 'count.positive?` with `present?`.

```ruby
require 'active_hash'

end

20000.times do
  Country.create
end

puts 'count.positive?'
before = Time.now
Country.count.positive?
after = Time.now
puts after - before

puts 'present?'
before = Time.now
Country.present?
after = Time.now
puts after - before

$ bundle exec ruby test.rb
count.positive?
7.0e-06
present?
3.0e-06
```

ise-tang:
I changed author email in each commits, so I did force push.

ise-tang:
@kbrock Could you do review again?

kbrock (approved):

PR #265, opened 2022-10-04
Relation: Make `where` return a fresh relation to avoid leaking scopes

pfeiffer:
This PR fixes issue #257 where scopes are leaking. The issue can lead to wildly unexpected behavior.

Previously calling `.where(..)` would mutate the original relation.

#### Before this change
```ruby
expect(countries.size).to eq 2
expect(countries.where(name: "US").size).to eq 1
expect(countries.size).to eq 2 # ❌ Fails as previous line mutated the `countries` object
```

#### After this change
```ruby
expect(countries.size).to eq 2
expect(countries.where(name: "US").size).to eq 1
expect(countries.size).to eq 2 # ✅ Passes as the previous line returned a new relation keeping the original
```

The change also makes the "dirty tracking" in `Relation` unnecessary as Relations are immutable.

### DIFF
--- a/lib/active_hash/relation.rb
+++ b/lib/active_hash/relation.rb
@@ -11,16 +11,14 @@ module ActiveHash
       self.klass = klass
       self.all_records = all_records
       self.query_hash = query_hash
-      self.records_dirty = false
-      self
     end
 
     def where(query_hash = :chain)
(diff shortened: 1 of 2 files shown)


pfeiffer:
See also the #268 as an alternative which addresses the root of the issue.

kbrock:
Going with #268

PR #266, opened 2022-10-04
Bugfix: Ensure that using `id: [..]` as condition returns unique records

pfeiffer:
This PR fixes an issue where only unique records will be returned when using `where(id: ..)`. Previously, if the same ID was provided multiple times in the `id` condition, it would be returned multiple times.

This PR fixes the issue and now mimics the way ActiveRecord work. It also streamlines so that `query_hash` always has symbols as keys. We can use `deep_symbolize_keys` for this as there's already a dependency on ActiveSupport.

### Before
```ruby
Country.where(id: [1, 2, 1])
# => 3 records; the Country with ID=1 is returned twice
```

### After
```ruby
Country.where(id: [1, 2, 1])
# => 2 records as expected
```

### DIFF
--- a/lib/active_hash/relation.rb
+++ b/lib/active_hash/relation.rb
@@ -10,7 +10,7 @@ module ActiveHash
     def initialize(klass, all_records, query_hash = nil)
       self.klass = klass
       self.all_records = all_records
-      self.query_hash = query_hash
+      self.query_hash = query_hash.deep_symbolize_keys
       self.records_dirty = false
       self
     end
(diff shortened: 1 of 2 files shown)


pfeiffer:
See also the #268 as an alternative which addresses the root of the issue.

kbrock:
At this moment, I am leaning towards #268 
Thanks again for providing this options (more surgical) and the other one (more like an atomic bomb ;) )

kbrock:
> so that query_hash always has symbols as keys

I had just read #196 and the thoughts around converting all keys to strings.
I'm guessing you don't care which camp this code is in, as long as it is consistent.

pfeiffer:
> I had just read #196 and the thoughts around converting all keys to strings.
> I'm guessing you don't care which camp this code is in, as long as it is consistent.

It's actually two totally different cases; for `query_hash` where we store the _conditions_ of the query, the type of keys in the hash is not that important, but I do think it's important to be consistent and not allow both strings *and* symbols as keys, but one of them.

For #196 I'm mentioning that __attributes__ should be stored with strings as keys, as this is what ActiveModel concerns expect and use for instance when serializing with `as_json`

Please note that in #268 also fixes the issue in this PR and supersedes that. This PR is the "surgical" one fixing the bug, while #268 addresses the overall issue (and a lot more) giving a better structure going forward IMO.

kbrock:
Going with #268

Issue #73, opened 2013-10-01
make `belongs_to` work again on AR?


zilkey:
```
class Country < ActiveHash::Base
end

class City < ActiveRecord::Base
  belongs_to :country # => this doesn't work, it needs to be...
  belongs_to_active_hash :country
end
```

I'm thinking to make this work, we could overwrite `belongs_to` on ActiveRecord::Base, make it check to see if we are attempting to belong_to an ActiveHash object, and if so,  just call `belongs_to_active_hash`.

PR #270, opened 2022-12-11
Avoid interfering with AR's belongs_to arguments.

koyo-miyamura:
# Background
In active_hash v3.1.0, `ActiveHash::Associations::ActiveRecordExtensions#belongs_to` copies arguments and calls `super` with no arguments.
It doesn't interfere with AR's `belongs_to` arguments.

https://github.com/active-hash/active_hash/blob/d952d67a0a52ee78f899366925757a79e4de6c37/lib/associations/associations.rb#L6-L24

However, after active_hash v3.1.1, it doesn't copy arguments and updates `options` from arguments and calls `super` with no arguments.
It interferes with AR's `belongs_to` arguments (please reference rspec in this PR).

# Changes
In this PR, avoid `ActiveHash::Associations::ActiveRecordExtensions#belongs_to` interfering with AR's `belongs_to` arguments.
(In addition, refactored the error handling as minor fix.)

### DIFF
--- a/lib/associations/associations.rb
+++ b/lib/associations/associations.rb
@@ -4,16 +4,16 @@ module ActiveHash
     module ActiveRecordExtensions
 
       def belongs_to(name, scope = nil, **options)
-        options = {:class_name => name.to_s.camelize }.merge(options)
+        klass_name = options.key?(:class_name) ? options[:class_name] : name.to_s.camelize
         klass =
           begin
-            options[:class_name].constantize
-          rescue
-            nil
-          rescue LoadError
+            klass_name.constantize
+          rescue StandardError, LoadError
             nil
           end
+
         if klass && klass < ActiveHash::Base
+          options = { class_name: klass_name }.merge(options)
           belongs_to_active_hash(name, options)
         else
           super
--- a/spec/associations/active_record_extensions_spec.rb
+++ b/spec/associations/active_record_extensions_spec.rb
@@ -166,6 +166,13 @@ unless SKIP_ACTIVE_RECORD
           expect(school.country).to eq(nil)
         end
 
+        it "doesn't interfere with AR's belongs_to arguments" do
+          allow(ActiveRecord::Base).to receive(:belongs_to).with(:country, nil)
+          allow(ActiveRecord::Base).to receive(:belongs_to).with(:country, nil, {})
+
+          School.belongs_to :country
+        end
+
         it "doesn't interfere w/ ActiveRecord's polymorphism" do
           School.belongs_to :locateable, :polymorphic => true
           school = School.new


koyo-miyamura:
@kbrock 
Thanks to launch CI ! 
I fixed to pass test in ruby 2.7. Please run again at your leisure.

spec/associations/active_record_extensions_spec.rb
```diff
@@ -166,6 +166,12 @@ class Book < ActiveRecord::Base
           expect(school.country).to eq(nil)
         end
 
+        it "doesn't interfere with AR's belongs_to arguments" do
+          allow(ActiveRecord::Base).to receive(:belongs_to).and_call_original
+          School.belongs_to :country
+          expect(ActiveRecord::Base).to have_received(:belongs_to).with(:country, nil)
+        end
+
```
koyo-miyamura, line 174: [FYI]
Before the fix, the rspec case failed as follows.
```ruby
$ bundle exec rspec spec/associations/active_record_extensions_spec.rb:149
Run options: include {:locations=>{"./spec/associations/active_record_extensions_spec.rb"=>[149]}}
.F....

Failures:

  1) ActiveHash::Base active record extensions ActiveHash::Associations::ActiveRecordExtensions#belongs_to doesn't interfere with AR's belongs_to arguments
     Failure/Error: expect(ActiveRecord::Base).to have_received(:belongs_to).with(:country, nil)
     
       #<ActiveRecord::Base (class)> received :belongs_to with unexpected arguments
         expected: (:country, nil)
              got: (:country, nil, {:class_name=>"Country"})
       Diff:
       @@ -1 +1 @@
       -[:country, nil]
       +[:country, nil, {:class_name=>"Country"}]
       
     # ./spec/associations/active_record_extensions_spec.rb:172:in `block (4 levels) in <top (required)>'

Finished in 0.17329 seconds (files took 0.32334 seconds to load)
6 examples, 1 failure

Failed examples:

rspec ./spec/associations/active_record_extensions_spec.rb:169 # ActiveHash::Base active record extensions ActiveHash::Associations::ActiveRecordExtensions#belongs_to doesn't interfere with AR's belongs_to arguments
```

kbrock: adding in the parameter `any` is your friend

koyo-miyamura: @kbrock 
Thank you for your advice! 
I will work on, please wait for a while.

kbrock:
this is going to be tricky.
different versions of active record have different interfaces into this method.

The original implementation #73 of this behavior used `super` like it is now
Then to work around some keyword issues, we modified the `super` call in #251
Then we went back to `super` in #255

So it looks like we have always been modifying `options`. But I can appreciate your change and it looks good.

I don't think checking the parameters coming into `belongs_to` is the best way to test. Possibly just testing that passing `:class_name` works?

To be honest, I'm not totally sure why you are trying to change this behavior. Maybe writing a test that is protecting you from this behavior may be the best bet?

spec/associations/active_record_extensions_spec.rb
```diff
@@ -166,6 +166,13 @@ class Book < ActiveRecord::Base
           expect(school.country).to eq(nil)
         end
 
+        it "doesn't interfere with AR's belongs_to arguments" do
+          allow(ActiveRecord::Base).to receive(:belongs_to).with(:country, nil)
+          allow(ActiveRecord::Base).to receive(:belongs_to).with(:country, nil, {})
+
+          School.belongs_to :country
+        end
```
koyo-miyamura, line 174: [FYI]
Before the fix, the rspec case failed in ruby 3.1.3 as follows.

```
$ BUNDLE_GEMFILE="gemfiles/rails_6.1.gemfile" bundle exec rspec
Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_checker)' has been deprecated. Please call `DidYouMean.correct_error(error_name, spell_checker)' instead.
.......................................................................................................................................................................................................................................................................................................F............................................

Failures:

  1) ActiveHash::Base active record extensions ActiveHash::Associations::ActiveRecordExtensions#belongs_to doesn't interfere with AR's belongs_to arguments
     Failure/Error: super
     
       #<ActiveRecord::Base (class)> received :belongs_to with unexpected arguments
         expected: (:country, nil, {}) (options hash)
              got: (:country, nil, {:class_name=>"Country"}) (keyword arguments)
       Diff:
       @@ -1 +1 @@
       -[:country, nil, {}]
       +[:country, nil, {:class_name=>"Country"}]
       
        Please stub a default value first if message might be received with other args as well. 
     # ./lib/associations/associations.rb:19:in `belongs_to'
     # ./spec/associations/active_record_extensions_spec.rb:173:in `block (4 levels) in <top (required)>'

Finished in 0.48655 seconds (files took 0.31712 seconds to load)
340 examples, 1 failure

Failed examples:

rspec ./spec/associations/active_record_extensions_spec.rb:169 # ActiveHash::Base active record extensions ActiveHash::Associations::ActiveRecordExtensions#belongs_to doesn't interfere with AR's belongs_to arguments
```

spec/associations/active_record_extensions_spec.rb
```diff
@@ -166,6 +166,13 @@ class Book < ActiveRecord::Base
           expect(school.country).to eq(nil)
         end
 
+        it "doesn't interfere with AR's belongs_to arguments" do
+          allow(ActiveRecord::Base).to receive(:belongs_to).with(:country, nil)
+          allow(ActiveRecord::Base).to receive(:belongs_to).with(:country, nil, {})
```
koyo-miyamura, line 171: There are differences between ruby 2.x and 3.x when keyword argument is passed.

It perhaps is caused by this issue.

- https://github.com/rspec/rspec-mocks/issues/1492

Therefore, to avoid this, I use two `allow(ActiveRecord::Base).to receive(:belongs_to).with` method.

It's enough to verify that it doesn't interfere with AR's `belongs_to` arguments, so the third argument can be `{}` or missing.

koyo-miyamura:
@kbrock 

Thank you for your advice and analyzing my PR.

> To be honest, I'm not totally sure why you are trying to change this behavior. Maybe writing a test that is protecting you from this behavior may be the best bet?

No wonder you think so because the problem we are facing now is complicated. 
Let me try to explain.

In our product, we also override AR's `belongs_to`, and the order of method call is after `ActiveHash::Associations::ActiveRecordExtensions#belongs_to`.

Therefore, `ActiveHash::Associations::ActiveRecordExtensions#belongs_to` adding `:class_name` to `options` causes trouble because our method can't distinguish whether `options[:class_name]` is originally included or not.

That's why I created this PR.

koyo-miyamura:
@kbrock 

Sorry for waiting you, the fix is ready, please review.

koyo-miyamura:
@kbrock 
Can you review this PR ?

lib/associations/associations.rb
```diff
@@ -4,16 +4,16 @@ module Associations
...
-        options = {:class_name => name.to_s.camelize }.merge(options)
+        klass_name = options.key?(:class_name) ? options[:class_name] : name.to_s.camelize
         klass =
           begin
-            options[:class_name].constantize
-          rescue
-            nil
-          rescue LoadError
+            klass_name.constantize
+          rescue StandardError, LoadError
```
kbrock, line 11: @koyo-miyamura Would you document why you need to add a `StandardError` here?

kbrock (approved):
Thanks for the explanation on why you are making this change.
It makes the intent that much easier.

Also, I like the way you handled the double `allow` for the various ruby kwargs cases.

lib/associations/associations.rb
koyo-miyamura: @kbrock 
I'm sorry but I cannot document this, because I only refactored this code.

Original code also handle `StandardError` but I don't explain why.

PR #271, opened 2023-01-16
Bugfix: Allow chaining with same attribute

gastronleroux:
Expected behavior:
```ruby
Country.where(id: 1).where(id: 2) # []
```
Current behavior:
```ruby
Country.where(id: 1).where(id: 2) # [#< Country id=2>]

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -43,7 +43,7 @@ module ActiveHash
           options.present? && match_options?(record, options)
         end
 
-        ActiveHash::Relation.new(@scope.klass, filtered_records, {})
+        ActiveHash::Relation.new(@scope.klass, filtered_records, [])
       end
 
       def match_options?(record, options)
@@ -205,7 +205,7 @@ module ActiveHash
       end
 
       def all(options = {})
-        ActiveHash::Relation.new(self, @records || [], options[:conditions] || {})
+        ActiveHash::Relation.new(self, @records || [], options[:conditions].to_a)
       end
 
       delegate :where, :find, :find_by, :find_by!, :find_by_id, :count, :pluck, :ids, :pick, :first, :last, :order, to: :all
--- a/lib/active_hash/relation.rb
+++ b/lib/active_hash/relation.rb
@@ -7,10 +7,10 @@ module ActiveHash
     delegate :empty?, :length, :first, :second, :third, :last, to: :records
     delegate :sample, to: :records
 
-    def initialize(klass, all_records, query_hash = nil)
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: lib/active_hash/base.rb [hunk 1]
```diff
@@ -205,7 +205,7 @@ def create!(attributes = {})
       end
 
       def all(options = {})
-        ActiveHash::Relation.new(self, @records || [], options[:conditions] || {})
+        ActiveHash::Relation.new(self, @records || [], options[:conditions].to_a || [])
```
kbrock, line 208: `nil.to_a == []` so the last `|| []` is no longer needed

spec/active_hash/base_spec.rb
```diff
@@ -307,15 +307,20 @@ class Region < ActiveHash::Base
...
 
       expect(where_relation.length).to eq 2
       expect(where_relation.map(&:id)).to eq([1, 2])
-      chained_where_relation = where_relation.where(name: 'US')
+      chained_where_relation = where_relation.where(name: "US")
       expect(chained_where_relation.length).to eq 1
       expect(chained_where_relation.map(&:id)).to eq([1])
     end
+
+    it "is chainable with same attribute" do
```
kbrock, line 320: nice test

lib/active_hash/base.rb
gastronleroux: Good catch!

lib/active_hash/relation.rb
```diff
@@ -19,7 +19,7 @@ def where(query_hash = :chain)
       return ActiveHash::Base::WhereChain.new(self) if query_hash == :chain
 
       self.records_dirty = true unless query_hash.nil? || query_hash.keys.empty?
-      self.query_hash.merge!(query_hash || {})
+      self.query_list.concat(query_hash&.to_a || [])
```
kbrock, line 22: so this is the main bug that this PR is addressing.

The merge will drop the existing key.

gastronleroux: Exactly

kbrock, line 22: Does it make sense to merge these two?

```ruby
unless query_hash.blank?
  self.records_dirty = true
  query_list.concat(query_hash.to_a)
end
```

kbrock (commented):
I admit that I am sad to see the hash go away and see the code move towards an array.

It only really shows up as a disadvantage when we are doing a search by id. So I guess it is no big deal.

I just have the one request for update which is minor
Please tweak that and squash your commits and we can merge this.

kbrock:

PR #272, opened 2023-01-16
Rails 7 support

flavorjones:
Alternative implementation to #267 that doesn't monkeypatch ActiveRecord.

This implementation subclasses `ActiveRecord::Reflection::BelongsToReflection` and overrides the `compute_class` method to allow ActiveHash associations.

I've also cherry-picked some relevant commits from #267.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -16,8 +16,7 @@ jobs:
         # | 6.1   |           |          |   2.5   |     |
         # | 7.0   |           |          |   2.7   |     |
         ruby: ['3.0', '3.1']
-        gemfile: ['gemfiles/rails_6.1.gemfile']
-          #   gemfile: 'gemfiles/rails_7.0.gemfile'
+        gemfile: ['gemfiles/rails_6.1.gemfile', 'gemfiles/rails_7.0.gemfile']
         include:
         - ruby: '2.4'
           gemfile: 'gemfiles/rails_5.0.gemfile'
--- a/lib/associations/associations.rb
+++ b/lib/associations/associations.rb
@@ -49,13 +49,20 @@ module ActiveHash
         end
 
         if ActiveRecord::Reflection.respond_to?(:create)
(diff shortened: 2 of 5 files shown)


kbrock:
@flavorjones This looks great. A few thoughts:

- Will this require developers to mixin this added module everywhere for this functionality?
- If so, we need to make sure this is a major version change with incompatibility note.
- Do we need to document this change so it is easy for developers to fix when upgrading to 7.0?

(as with everywhere in rubyland. thanks for the contribution)

kbrock (approved):
LGTM. just minor logistics questions

flavorjones:
@kbrock Thanks for the review and for asking these questions.

This should "just work" for folks who have followed the README and are extending their AR classes with `ActiveHash::Associations::ActiveRecordExtensions` (see [this section in the README](https://github.com/active-hash/active_hash#referencing-activehash-objects-from-activerecord-associations)).

However, it looks like not _all_ the tests were following that advice but were still passing with Rails 6 (see the change in this PR to `spec/active_hash/base_spec.rb`), which seems suspicious and makes me believe there may be apps out there which are "working" but will break with Rails 7. Maybe I'm not fully understanding Active Hash and how it's supposed to be used? I could use some advice, honestly.

flavorjones:
Looks like the ActiveRecordExtensions module was added in v0.9.6 (see 153c58ec) but wasn't added to this test (probably because it wasn't failing).

kbrock:
Thanks.
I thought I remembered this module from a number of years ago.

Ok, I change my view from breaking change to bug fix.

flavorjones:
@kbrock I just saw now that I still had a "wip" message in that last commit, urk, sorry, I should have fixed that. Candidly, I did not expect that code to pass on every Rails version and assumed I'd be coming back to clean something up. :shrug:

kbrock:
@flavorjones ugh. very sorry that I did not see that. ping me to get that through

PR #277, opened 2023-05-08
fix: provide a cleaner method missing error

rgarver:
I hit a weird error that ultimately was my fault, but it exposed a use-before-initialize bug in `ActiveHash::Relation#method_missing` on the `.scopes` class_attribute. I changed the class_attribute to a class method with a memoized class variable that is lazily initialized on reference.

This is a low-impact addition.

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -21,7 +21,7 @@ module ActiveHash
   end
 
   class Base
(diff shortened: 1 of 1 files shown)


kbrock:
Will this be solved via https://github.com/active-hash/active_hash/pull/278

would prefer to not introduce the `scopes` variable if possible.

rgarver:
Indeed, much simpler. Good catch.

PR #281, opened 2023-06-07
Condition explicitly uses attribute values and not public_send

### LINKED ISSUE #280: 3.2.0: Regression in `.where`?
Hiya!

SYSTEM INFO:
Rails: 6.1.7.1
Ruby: 3.0.4

Old version of ActiveHash: 3.1.1
New version: 3.2.0

I'm working on updating our version of ActiveHash to 3.2 from 3.1.1. We use this gem extensively because it's great, and it's usually this (example is not direct from our app, but is structured the same way):

```
class Disposition < ActiveHash::Base
  include ActiveHash::Enum

  enum_accessor :name
  self.data = [
    {
      id: 1,
      name: :a_disposition,
      display: true,
    },
    {
      id: 2,
      name: :another_disposition,
      display: false,
    },
  ]
end
```

We have a lot of code that uses `.where` to filter our dictionaries: `Disposition.where(display: true)`. This is still locating the correct records, but some of the changes to `.where` in 3.2.0 seem to have broken our ability to call `.map(&:name)` on those records. I'm expecting to see an array of names associated with our dictionary, and am getting this as output: 
```
#<Disposition:0x00000001418dd4b0>#<Disposition:0x00000001418dcab0>#<Disposition:0x00000001418dc5b0>#<Disposition:0x00000001418d7ec0>#<Disposition:0x00000001418d7858>#<Disposition:0x00000001418d71c8>#<Disposition:0x00000001418d6c50>=> []
```

I can get the expected list by going via `.all` or calling `.all_records` on the output, but I was surprised to see this breaking on us during my upgrade.

Is this a bug? Did something change here?

flavorjones:
fix: Condition explicitly uses attribute values and not public_send

Using public_send caused an issue when attribute names shadowed method names.

Closes #280

### DIFF
--- a/lib/active_hash/condition.rb
+++ b/lib/active_hash/condition.rb
@@ -19,7 +19,7 @@ class ActiveHash::Relation::Condition
       expectation_method = inverted ? :any? : :all?
 
       constraints.send(expectation_method) do |attribute, expected|
-        value = record.public_send(attribute)
+        value = record.read_attribute(attribute)
 
         matches_value?(value, expected)
       end
--- a/spec/active_hash/relation_spec.rb
+++ b/spec/active_hash/relation_spec.rb
@@ -50,4 +50,25 @@ RSpec.describe ActiveHash::Relation do
       expect(array.size).to eq(2)
     end
   end
+
+  describe "colliding methods https://github.com/active-hash/active_hash/issues/280" do
+    it "should handle attributes named after existing methods" do
+      klass = Class.new(ActiveHash::Base) do
+        self.data = [
+          {
+            id: 1,
+            name: "Aaa",
+            display: true,
+          },
+          {
+            id: 2,
+            name: "Bbb",
+            display: false,
+          },
+        ]
+      end
+
+      expect(klass.where(display: true).length).to eq(1)
+    end
+  end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # active_hash Changelog
 
+## next - unreleased
+
+- Fix relation matching when attribute name collides with a method. [#281](https://github.com/active-hash/active_hash/pull/281) @flavorjones
+
+
 ## Version [3.2.0] - <sub><sup>2022-07-14</sup></sub>
 
 - Add Ruby 3.2 to the CI matrix [#275](https://github.com/active-hash/active_hash/pull/275) @petergoldstein


flavorjones:
@kbrock When you get some time, would love any feedback you have on this potential fix. (No rush, I know you've been busy!)

spec/active_hash/relation_spec.rb
```diff
@@ -50,4 +50,24 @@
...
+            display: true,
+          },
+          {
+            id: 2,
+            name: "Bbb",
+            display: false,
+          },
+        ]
+      end
+      expect(klass.where(display: true).length).to eq(1)
```
kbrock, line 70: maybe also add:

```ruby
   def name
     self[:name] + "!"
   end
```

```ruby
expect(klass.where(name: 'Bbb!').length).to eq(1)
```

lib/active_hash/condition.rb
```diff
@@ -19,7 +19,7 @@ def matches?(record)
       expectation_method = inverted ? :any? : :all?
 
       constraints.send(expectation_method) do |attribute, expected|
-        value = record.public_send(attribute)
+        value = record[attribute]
```
kbrock, line 22:
```suggestion
        value = record.read_attribute(attribute)
```

I think this is basically the same thing.
so I have no issues merging this without this change.

kbrock (changes requested):
Wow, great work/find.

The good old column name conflicts with Object namespace trick.

I currently have one concern (as you can see from the comments):

Will this break custom accessor methods?

`ActiveRecord#read_attribute()` has a pretty serious implementation that detects custom accessor methods. Think their `[]` calls `read_attribute()`

We have tests to handle custom attribute accessor methods, so if those passed with this test, I guess that means this code change is good?

You know, I don't think rails would work with this case either. Isn't that the whole reason why `Object#id` changed to `Object#object_id`?

Having said that, I'm leaning towards merging this.

@pfeiffer Let us know if you can see anything wrong with this.

flavorjones:
Yeeaaaah, @kbrock good catch, this _does_ break custom accessor methods. Gonna give this a good think.

flavorjones:
OK, taking a step back for a moment ... the code that calls `public_send` was added in aa512c4d as part of #268. Before that commit, `where` did not support arbitrary method definitions.

Here are the tests I'm running:

```ruby
  describe "colliding methods" do
    klass = Class.new(ActiveHash::Base) do
      self.data = [
        {
          id: 1,
          name: "Aaa",
          display: true,
        },
        {
          id: 2,
          name: "Bbb",
          display: false,
        },
      ]

      def name
        self[:name] + "!"
      end

      def not_a_real_attribute
        self[:name]
      end
    end

    it "should handle attributes named after existing methods" do
      expect(klass.where(display: true).length).to eq(1)
    end

    it "should handle redefined attributes" do
      expect(klass.where(name: 'Bbb!').length).to eq(1)
    end

    it "should handle arbitrary methods" do
      expect(klass.where(not_a_real_attribute: 'Bbb').length).to eq(1)
    end
  end
```

On e58f3ff (**before** the Relation overhaul was merged), the results are:

```
Failures:

  1) ActiveHash::Relation colliding methods should handle redefined attributes
     Failure/Error: expect(klass.where(name: 'Bbb!').length).to eq(1)
     
       expected: 1
            got: 0
     
       (compared using ==)
     # ./spec/active_hash/relation_spec.rb:83:in `block (3 levels) in <top (required)>'

  2) ActiveHash::Relation colliding methods should handle arbitrary methods
     Failure/Error: expect(klass.where(not_a_real_attribute: 'Bbb').length).to eq(1)
     
       expected: 1
            got: 0
     
       (compared using ==)
     # ./spec/active_hash/relation_spec.rb:87:in `block (3 levels) in <top (required)>'
```

and *after* #268 is merged:

```
Failures:

  1) ActiveHash::Relation colliding methods should handle attributes named after existing methods
     Failure/Error: expect(klass.where(display: true).length).to eq(1)
     
       expected: 1
            got: 0
     
       (compared using ==)
     # ./spec/active_hash/relation_spec.rb:79:in `block (3 levels) in <top (required)>'

```

So my thinking is that it's OK to break support for arbitrary custom accessor methods, since:

- it was only introduced (not intentionally) in v3.2.0, and presents the bug in #280 
- Active Record doesn't support custom accessor methods with `where`, meaning that potential migrations may break behavior relied upon in Active Hash

I'd like to suggest that we merge this PR and cut a release with a clear CHANGELOG description.

spec/active_hash/relation_spec.rb
flavorjones: AFAICT this isn't something Active Record supports (see note below).

kbrock (approved):

kbrock:
So I ran the same test against active record.
This works great.

Thanks for making the `read_attribute` changes.

pfeiffer:
Hey, a bit late to this PR and the request for comments. I think using `read_attribute` makes sense here, so great change!

PR #286, opened 2023-08-02
require reflection_extensions when ActiveRecordExtensions is extended

iberianpig:
**Problem**

Previously, unless `include ActiveHash::Associations` was executed, the override to `compute_class` when setting `belongs_to_active_hash` would not be performed. This is an unintended behavior.

When `compute_class` is called in Rails 7, the following error occurs:

```
ArgumentError: Rails couldn't find a valid model for Rank association.
Please provide the :class_name option on the association declaration.
If :class_name is already provided, make sure it's an ActiveRecord::Base subclass.
```

**Solution**

This pull request proposes a change to `require "reflection_extensions"` when `extend ActiveHash::Associations::ActiveRecordExtensions` is called instead of `include ActiveHash::Associations`

The reflection is defined only when `extend ActiveHash::Associations::ActiveRecordExtensions`.
As a result, the `compute_class` is properly overridden.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## next - unreleased
 
 - Fix relation matching when attribute name collides with a method. [#281](https://github.com/active-hash/active_hash/pull/281) @flavorjones
+- Fix association reflection in applications that don't use ActiveHash::Associations. [#286](https://github.com/active-hash/active_hash/pull/286) @iberianpig
 
 
 ## Version [3.2.0] - <sub><sup>2023-05-06</sup></sub>
--- a/lib/associations/associations.rb
+++ b/lib/associations/associations.rb
@@ -3,6 +3,10 @@ module ActiveHash
 
     module ActiveRecordExtensions
 
+      def self.extended(base)
+        require_relative 'reflection_extensions'
+      end
+
       def belongs_to(name, scope = nil, **options)
         klass_name = options.key?(:class_name) ? options[:class_name] : name.to_s.camelize
         klass =
@@ -92,7 +96,6 @@ module ActiveHash
     end
 
     def self.included(base)
-      require_relative "reflection_extensions"
       base.extend Methods
     end
 


flavorjones:
@iberianpig Thank you for pointing out this issue and for suggesting a fix!

Can you help me reproduce the issue that you're seeing? I'm not entirely sure I understand what you're seeing.

iberianpig:
I want to make the situation easier to understand, so I will create a reproduction procedure. Please wait for a while.

iberianpig:
I've created a reproduction procedure. The following code can be used to reproduce the issue.


```ruby
begin
  require "bundler/inline"
rescue LoadError => e
  $stderr.puts "Bundler version 1.10 or later is required. Please update your Bundler"
  raise e
end


gemfile = File.read(File.expand_path("Gemfile", __dir__)).strip
sqlite3_version = gemfile.lines.find { |line| line.include?('sqlite3') }
  .then { |sqlite3_line| sqlite3_line.strip.split(',').last.delete('"') }

gemfile(true) do
  source "https://rubygems.org"
  gem "activerecord", "~> 7.0.0"
  gem "sqlite3", sqlite3_version
  gem "debug"
  gem "active_hash" # original
  # gem "active_hash", github: "iberianpig/active_hash", branch: "fix/hook-for-reflection-extensions" # fixed version
end

require "active_record"
require "active_hash"
require "minitest/autorun"
require "logger"

# This connection will do for database-independent bug reports.
ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")

  create_table :People, force: true do |t|
    t.integer :country_id
  end
end

  self.data = [
    {:id => 1, :name => "US"},
    {:id => 2, :name => "Canada"}
  ]
end

class Person < ActiveRecord::Base
  belongs_to_active_hash :country, :shortcuts => [:name]
end

class BugTest < Minitest::Test
  def test_association_stuff
    john = Person.new
    john.country_name = "US"
    # Is the same as doing `john.country = Country.find_by_name("US")`
    john.country_name
    # Will return "US", and is the same as doing `john.country.try(:name)`
    assert_equal Country.find_by_name("US"), john.country

    # Should compute and return the Country class
    assert_equal Country, Person.reflect_on_all_associations.find { |a| a.name == :country }.klass
    # => eval error: Rails couldn't find a valid model for Country association. Please provide the :class_name option on the association declaration. If :class_name is already provided, make sure it's an ActiveRecord::Base subclass.
    #   /home/iberianpig/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/activerecord-7.0.6/lib/active_record/reflection.rb:436:in `compute_class'
    #   /home/iberianpig/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/activerecord-7.0.6/lib/active_record/reflection.rb:382:in `klass'
    #   (rdbg)/activehash_issue.rb:1:in `test_association_stuff'
  end
end
```

Please save this file as "issue.rb" and execute it by running `ruby issue.rb` .

Additionally, by uncommenting the line, you can test my correction.

```ruby
  # gem "active_hash", github: "iberianpig/active_hash", branch: "fix/hook-for-reflection-extensions" # fixed
```
The problem arises because the code for Rails 7.0 isn't being triggered during the reflection process.

flavorjones:
Ah! Right, that totally makes sense, and this isn't being caught by the test suite. This is a bug introduced (by me) in #272 that doesn't show up if _any class in the application_ is decorated with `include ActiveHash::Associations`.

So: I'm not sure how to test this, but this is the right fix.

flavorjones (approved):

PR #287, opened 2023-08-03
Fix (Relation): add safe navigator operator after scopes for respond_to_missing? and method_missing methods

Thornolf:
Hello ! 

In our attempt to move our project to Ruby 3.2.2 we encountered an error : `MethodError: undefined method 'key?' for nil:NilClass`, the origin of the error was from `relation.rb:173`, our class did not had any scopes and so it was sending this error. We had multiples ideas to fix the issue but the simplest one was to just add a safe navigator operator. What do you think about that ? I'm available if you need me to do any other modifications.

Thank you !

### DIFF
--- a/lib/active_hash/relation.rb
+++ b/lib/active_hash/relation.rb
@@ -164,13 +164,13 @@ module ActiveHash
     end
 
     def method_missing(method_name, *args)
-      return super unless klass.scopes.key?(method_name)
+      return super unless klass.scopes&.key?(method_name)
 
       instance_exec(*args, &klass.scopes[method_name])
     end
 
     def respond_to_missing?(method_name, include_private = false)
-      klass.scopes.key?(method_name) || super
+      klass.scopes&.key?(method_name) || super
     end
 
     private


flavorjones:
@Thornolf Thank you for pointing out this issue and suggesting a fix!

Would you mind adding a test case that demonstrates the issue, to prevent regressions?

Thornolf:
> @Thornolf Thank you for pointing out this issue and suggesting a fix!
> 
> Would you mind adding a test case that demonstrates the issue, to prevent regressions?

Sure ! I will do it as soon as possible :) Thank you for the review !

kbrock:
Thanks you @Thornolf

It looks like https://github.com/active-hash/active_hash/pull/278 was the same fix.
I just merged that not knowing that this was also available.

I am closing this one as duplicate, but please let me know if you have other concerns

Thornolf:
Oh indeed, sorry for the delay I was a bit busy at work 😅 I did not even saw the PR myself sorry for creating this duplicate ! Thank you for merging it 💪🏻

PR #292, opened 2023-10-24
Fix #where for hashes with string keys

usernam3:
✌️  Hey `active_hash` maintainers ✌️ Thank you for such great gem that makes development with ruby even more productive 🙇 

I noticed that the latest version (`v3.2.1`) has undocumented side-effect which affect the behaviour of one of the core features - search. To be specific the `#where` for hash with string keys stopped working (example: `ActiveHashModel.where("id" => 1)`).
It seems that transition from `Object#public_send ` to `ActiveHash::Base#read_attribute` (in #281) isn't backward-compatible as those methods have a bit different expectation towards arguments. [Object#public_send](https://apidock.com/ruby/Object/public_send#:~:text=When%20the%20method%20is%20identified%20by%20a%20string%2C%20the%20string%20is%20converted%20to%20a%20symbol) - converts arguments to symbols, `read_attribute` - not (while `attributes` internally are stored with symbolized keys).

<details>
  <summary>Small demo</summary>

```ruby
(rdbg) pp record
#<Category:0x000000012a0d7db8 @attributes={:id=>1, :name=>"qwerty"}>
(rdbg) pp record.read_attribute("id")
nil
(rdbg) pp record.public_send(:id)
1
(rdbg) pp record.attributes["id"]
nil
(rdbg) pp record.attributes[:id]
1
(rdbg) pp record.id
1
```  
  
</details>

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -411,7 +411,7 @@ module ActiveHash
     end
 
     def _read_attribute(key)
-      attributes[key]
+      attributes[key.to_sym]
     end
     alias_method :read_attribute, :_read_attribute
 
--- a/lib/active_hash/condition.rb
+++ b/lib/active_hash/condition.rb
@@ -19,7 +19,7 @@ class ActiveHash::Relation::Condition
       expectation_method = inverted ? :any? : :all?
 
       constraints.send(expectation_method) do |attribute, expected|
-        value = record.read_attribute(attribute.to_sym)
+        value = record.read_attribute(attribute)
 
         matches_value?(value, expected)
       end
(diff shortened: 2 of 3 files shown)


usernam3:
@kbrock Cool idea, thanks for suggesting it, it definitely makes things consistent.

Conversions are inside the `_read_attribute` now, initially I split it into two separate methods

like:
```ruby
    def _read_attribute(key)
      attributes[key.to_s]
    end

    def read_attribute(key)
      attributes[key.to_sym]
    end
```

But didn't find a reason to keep them both, as `attributes[key.to_s]` doesn't make much sense (?or does it?) as `attributes` have keys symbolized.

kbrock:
@usernam3 thanks for the change.

```ruby
vm.read_attribute(:name)
# => "abc"
vm.read_attribute("name")
=> "abc"
vm._read_attribute(:name)
# => nil
vm._read_attribute("name")
# => "abc"
```

Yes, technically `read_attribute` does some conversion and `_read_attribute` is more skeletal, but this works.

kbrock (approved):
Looking good

PR #296, opened 2023-12-02
support `has_many :through` associations

### LINKED ISSUE #239: The has_many :through Association does not get record
### Steps to reproduce

Run below code.

```ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem 'activerecord', '~> 6.1.6'
  gem 'active_hash', '~> 3.2.1'

  gem 'sqlite3'
end

require 'active_record'
require 'active_hash'

require 'minitest/autorun'
require 'logger'

ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: ':memory:')

  create_table :appointments, force: true do |t|
    t.references :physician
    t.references :patient
  end

  create_table :patients, force: true do |t|
  end
end

class Physician < ActiveHash::Base

  has_many :appointments
  has_many :patients, through: :appointments

  self.data = [
    {:id => 1, :name => "ikeda"},
    {:id => 2, :name => "sato"}
  ]
end

class Appointment < ActiveRecord::Base

  belongs_to :physician
  belongs_to :patient
end

class Patient < ActiveRecord::Base

  has_many :appointments
  has_many :physicians, through: :appointments
end

class TestAssociation < Minitest::Test
  def test_association
    physician = Physician.first
    patient = Patient.create!
    Appointment.create!(physician: physician, patient: patient)

    assert_equal [physician], patient.appointments.flat_map(&:physician) # works
    assert_equal [physician], patient.physicians # error
  end
end
```

### Expected behavior

The script above should produce no errors, and get physician.

### Actual behavior

> NoMethodError: undefined method `relation_delegate_class' for Physician:Class

### System configuration

**Ruby version**: 2.7.4
**ActiveHash version**: 3.1.0

updated:
**Ruby version**: 3.0.6
**ActiveHash version**: 3.2.1

updated by @kbrock to bring up to current version of ruby, rails, and active hash

flavorjones:
This PR closes #239.

The first commit removes unused (and broken?) code that I found while looking at the association code. It was part of e6a47bcc and should not have been present.

The second commit refactors the extensions spec a bit to minimize the test classes being created for each test; this makes the tests more explicit and speeds them up slightly.

The final commit introduces the test from #239 and overrides the `has_many` method in any class that includes `ActiveHash::Associations::ActiveRecordExtensions`. The new implementation will adopt new behavior if the `:through` option is present and it points at an `ActiveHash::Base` class:

- retrieving the join models (e.g., "Patient#appointments")
- calling the `:through` association name on each join model (e.g., "Appointment#physician")
- and returning the unique matching records

Otherwise `has_many` will just call `super`.

### DIFF
--- a/lib/associations/associations.rb
+++ b/lib/associations/associations.rb
@@ -55,9 +55,6 @@ module ActiveHash
         if ActiveRecord::Reflection.respond_to?(:create)
           if defined?(ActiveHash::Reflection::BelongsToReflection)
             reflection = ActiveHash::Reflection::BelongsToReflection.new(association_id.to_sym, nil, options, self)
-            if options[:through]
-              reflection = ActiveRecord::ThroughReflection.new(reflection)
-            end
           else
             reflection = ActiveRecord::Reflection.create(
               :belongs_to,
--- a/lib/associations/associations.rb
+++ b/lib/associations/associations.rb
@@ -2,9 +2,33 @@ module ActiveHash
   module Associations
 
     module ActiveRecordExtensions
-
       def self.extended(base)
         require_relative 'reflection_extensions'
+      end
+
+      def has_many(association_id, **options)
+        if options[:through]
+          klass_name = association_id.to_s.classify
+          klass =
+            begin
+              klass_name.constantize
+            rescue StandardError, LoadError
+              nil
+            end
+
+          if klass && klass < ActiveHash::Base
+            define_method(association_id) do
+              join_models = send(options[:through])
+              join_models.flat_map do |join_model|
+                join_model.send(association_id.to_s.singularize)
+              end.uniq
+            end
+
+            return
+          end
+        end
+
+        super
       end
 
       def belongs_to(name, scope = nil, **options)
@@ -117,6 +141,7 @@ module ActiveHash
             klass.where(foreign_key => primary_key_value)
           end
         end
+
         define_method("#{association_id.to_s.underscore.singularize}_ids") do
           public_send(association_id).map(&:id)
         end
@@ -140,7 +165,6 @@ module ActiveHash
       end
 
       def belongs_to(association_id, options = {})
-
         options = {
           :class_name => association_id.to_s.classify,
           :foreign_key => association_id.to_s.foreign_key,
@@ -156,7 +180,6 @@ module ActiveHash
         define_method("#{association_id}=") do |new_value|
           attributes[options[:foreign_key].to_sym] = new_value ? new_value.send(options[:primary_key]) : nil
         end
-
       end
     end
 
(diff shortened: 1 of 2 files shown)


spec/associations/active_record_extensions_spec.rb
```diff
@@ -3,65 +3,115 @@
...
     after do
-      Object.send :remove_const, :City
-      Object.send :remove_const, :Author
-      Object.send :remove_const, :Country
-      Object.send :remove_const, :School
-      Object.send :remove_const, :SchoolStatus
-      Object.send :remove_const, :Book
+      @ephemeral_classes.each do |klass_name|
+        Object.send :remove_const, klass_name
+      end
```
kbrock, line 108: Do we need to do more here?

On another project I needed to tell rails 6.1 to forget about an active record model that was cached. This crept up when using STI. Maybe we do not need to do this since we are not using STI here?

snipped I had to use, but again, this was a different use case and this code may be unnecessary.

```ruby
  # these descendants are stored in a cache in active support
  # this cleans out those values so future runs do not have bogus classes
  # this causes sporadic test failures.
  def cleanup_subclass(parent, children)
    tracker = ActiveSupport::DescendantsTracker.class_variable_get(:@@direct_descendants)[parent]
    tracker&.reject! { |child| children.include?(child) }
  end
```

lib/associations/associations.rb
```diff
@@ -2,11 +2,35 @@ module ActiveHash
...
       def self.extended(base)
         require_relative 'reflection_extensions'
       end
 
+      def has_many(association_id, **options)
+        if options[:through]
+          klass_name = association_id.to_s.classify
+          klass =
+            begin
+              klass_name.constantize
```
kbrock, line 14: This makes me nervous.
I believe has_many is executed within a lock since it is defining a class.

Think this issue explains the issue: https://github.com/rails/rails/issues/34310

In our projects, when we are defining a class, we work hard to not define a second one.
Not sure if this race condition still exists.

spec/associations/active_record_extensions_spec.rb
kbrock: Yea, and my code referenced an internal rails private variable. Nothing could go wrong with that :(

I really like this change though.
I suppose it improves the original and does not make it worse

lib/associations/associations.rb
kbrock: ok, this is the same code path as our belongs_to fix.

I wasn't able to follow the association builder code to find an answer
looks like they do call safe_constantize (sparingly)

kbrock (approved):
@flavorjones the 2 issues I see are not really introduced by this PR but just propagation of existing issues.

If I find out how rails fixes this, then I'll introduce a fix later on. (I'm not hopeful though)

Lets merge since this is a great fix.

lib/associations/associations.rb
flavorjones: Hmm. I'll create a separate PR converting all our invocations to `safe_constantize`.

flavorjones: 🤦 I definitely did not read carefully enough. Apologies for missing your point, I get it now.

PR #297, opened 2023-12-06
prefer Active Support's String#safe_constantize to #constantize

flavorjones:
In #296, @kbrock pointed out that Rails uses `String#safe_constantize` when finding a `Class` from a class name.

Functionally I think this will be identical, so maybe there's no good reason to make the change other than "following Rails practices as closely as we can".


kbrock (approved):
Well, I haven't been able to determine when rails uses `constantize` and when they use `safe_constantize`

LGTM

PR #269, opened 2022-11-28
Fix #pluck method behavior when 3+ attributes specified

h6ah4i:
Hi. I noticed that ActiveHash's `#pluck` method behaves differently from ActiveRecord one. This pull request fixes the issue.

### Current behavior (ActiveHash v3.1.1)

```ruby
Category.pluck(:id, :name, :language)
=> [
  [[1, "US"], "English"],
  [[2, "Canada"], "English"],
  [[3, "Mexico"], "Spanish"]
]
```

### Expected behavior (same as ActiveRecord)

```ruby
Category.pluck(:id, :name, :language)
=> [
  [1, "US", "English"],
  [2, "Canada", "English"],
  [3, "Mexico", "Spanish"]
]
```

PR #299, opened 2024-02-06
pluck change from array access to using methods.

iberianpig:
Restore ActiveHash::Relation#pluck to execute the specified method on the record.

## Background:

- Through v3.1.x: Executed method names given as symbols.
- In v3.2.x: Change impacted data retrieval by directly referencing `attributes[key]`, affecting compatibility.

https://github.com/active-hash/active_hash/pull/269 broke functionality of gems like `enumerize` which rely on casting values accessed by column names.

## Reproduce

1. Save following file as issue.rb.
2. Please switch version with uncomment line of `gem "active_hash", "3.2.1" # fail test`.
3. Run in terminal `$ ruby issue.rb`

```ruby
#!/usr/bin/env ruby

begin
  require "bundler/inline"
rescue LoadError => e
  $stderr.puts "Bundler version 1.10 or later is required. Please update your Bundler"
  raise e
end

gemfile(true) do
  source "https://rubygems.org"
  gem 'enumerize'
  # gem "active_hash", "3.1.1" # pass test
  # gem "active_hash", "3.2.1" # fail test
  gem "active_hash", github: "active-hash/active_hash", ref: "6c51fa6d" # == master # fail test
  # gem "active_hash", github: "iberianpig/active_hash", branch: "fix/revert-pluck-behavior" # pass test
end

require "active_hash"
require "minitest/autorun"
require "logger"

  extend Enumerize
  enumerize :continent, in: { north_america: 1, south_america: 2, europe: 3, asia: 4, africa: 5, oceania: 6 }
  self.data = [
      {
        "id": 1,
        "name": "US",
        "custom_field_1": 1,
        "continent": 1
      },
      {
        "id": 2,
        "name": "Canada",
        "custom_field_2": 2,
        "continent": 1
      }
    ]
end
class BugTest < Minitest::Test
  def test_pluck_with_enumerize
    assert_equal "north_america", Country.find(1).continent  # continent is enumerized
    assert_equal [["US", "north_america"], ["Canada", "north_america"]], Country.all.pluck(:name, :continent)
  end
end
```
### 3.1.1
```
Fetching gem metadata from https://rubygems.org/.........
Resolving dependencies...
Run options: --seed 37444

# Running:

.

Finished in 0.000997s, 1002.6591 runs/s, 2005.3181 assertions/s.

1 runs, 2 assertions, 0 failures, 0 errors, 0 skips
```

### 3.2.1
```diff
Fetching gem metadata from https://rubygems.org/.........
Resolving dependencies...
Run options: --seed 14751

# Running:

F

Finished in 0.004705s, 212.5528 runs/s, 425.1055 assertions/s.

  1) Failure:
BugTest#test_pluck_with_enumerize [issue.rb:43]:
--- expected
+++ actual
@@ -1 +1 @@
-[["US", "north_america"], ["Canada", "north_america"]]
+[["US", 1], ["Canada", 1]]


1 runs, 2 assertions, 1 failures, 0 errors, 0 skips

```

### DIFF
--- a/lib/active_hash/relation.rb
+++ b/lib/active_hash/relation.rb
@@ -143,15 +143,10 @@ module ActiveHash
     end
 
     def pluck(*column_names)
-      symbolized_column_names = column_names.map(&:to_sym)
-
-      if symbolized_column_names.length == 1
-        column_name = symbolized_column_names.first
-        all.map { |record| record[column_name] }
+      if column_names.length == 1
+        all.map(&column_names.first.to_sym)
       else
-        all.map do |record|
-          symbolized_column_names.map { |column_name| record[column_name] }
-        end
+        column_names.map { |column_name| all.map(&column_name.to_sym) }.then { |values| :zip.to_proc.(*values) }
       end
     end
 
--- a/lib/active_hash/relation.rb
+++ b/lib/active_hash/relation.rb
@@ -144,10 +144,10 @@ module ActiveHash
 
     def pluck(*column_names)
       if column_names.length == 1
-        all.map(&column_names.first.to_sym)
+        column_name = column_names.first
+        all.map { |record| record.public_send(column_name) }
       else
-        # `tap with break` can be replaced with yield_self in Ruby 2.5 or then in Ruby 2.6
-        column_names.map { |column_name| all.map(&column_name.to_sym) }.tap { |values| break :zip.to_proc.(*values) }
+        all.map { |record| column_names.map { |column_name| record.public_send(column_name) } }
       end
     end
 
--- a/lib/active_hash/relation.rb
+++ b/lib/active_hash/relation.rb
@@ -146,7 +146,8 @@ module ActiveHash
       if column_names.length == 1
         all.map(&column_names.first.to_sym)
       else
-        column_names.map { |column_name| all.map(&column_name.to_sym) }.yield_self { |values| :zip.to_proc.(*values) }
+        # `tap with break` can be replaced with yield_self in Ruby 2.5 or then in Ruby 2.6
+        column_names.map { |column_name| all.map(&column_name.to_sym) }.tap { |values| break :zip.to_proc.(*values) }
       end
     end
 
--- a/lib/active_hash/relation.rb
+++ b/lib/active_hash/relation.rb
@@ -146,7 +146,7 @@ module ActiveHash
       if column_names.length == 1
         all.map(&column_names.first.to_sym)
       else
-        column_names.map { |column_name| all.map(&column_name.to_sym) }.then { |values| :zip.to_proc.(*values) }
+        column_names.map { |column_name| all.map(&column_name.to_sym) }.yield_self { |values| :zip.to_proc.(*values) }
       end
     end
 
--- a/spec/active_hash/base_spec.rb
+++ b/spec/active_hash/base_spec.rb
@@ -599,6 +599,30 @@ describe ActiveHash, "Base" do
     it "returns an Array of attribute values" do
       expect(Country.pluck(:id)).to match_array([1, 2, 3])
     end
+
+    context 'with the same field name and method name' do
+      before do
+        class CountryWithContinent < ActiveHash::Base
+          self.data = [
+            {:id => 1, :name => "US", continent: 1},
+            {:id => 2, :name => "Canada", continent: 1},
+            {:id => 3, :name => "Mexico", continent: 1},
+            {:id => 4, :name => "Brazil", continent: 2}
+          ]
+
+          # behave like ActiveRecord Enum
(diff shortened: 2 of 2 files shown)


flavorjones:
@iberianpig see #300 for the Rails 7.1.3 fix

kbrock:
kicking now that 300 was merged

kbrock:
looking like the tests introduced by 299 are still passing.
We had introduced a few other changes around a symbol or string column access, so maybe the changes in 299 are no longer needed?

iberianpig:
@kbrock Please confirm Reproduce section in this issue. You can reproduce this issue  in your terminal.


> looking like the tests introduced by 299 are still passing.

I rebased and pushed this branch, then CI is all passed.
However it's still failing with `$ ruby issue.rb` .

```diff
  1) Failure:
BugTest#test_pluck_with_enumerize [issue.rb:44]:
--- expected
+++ actual
@@ -1 +1 @@
-[["US", "north_america"], ["Canada", "north_america"]]
+[["US", 1], ["Canada", 1]]


1 runs, 2 assertions, 1 failures, 0 errors, 0 skips
```

kbrock:
@iberianpig Sorry, I was not clear. Let me try again.

Good job. after this change, the tests introduced by #269 are still passing.
This makes me feel that the code change introduced in #269 is no longer needed.

iberianpig:
Thank you for your feedback.
I've already add tests for `#pluck` that it does not rely on enumerize. Could you please take a moment to review them in This PR?

lib/active_hash/relation.rb
```diff
@@ -143,15 +143,11 @@ def size
...
-        column_name = symbolized_column_names.first
-        all.map { |record| record[column_name] }
+      if column_names.length == 1
+        all.map(&column_names.first.to_sym)
       else
-        all.map do |record|
-          symbolized_column_names.map { |column_name| record[column_name] }
-        end
+        # `tap with break` can be replaced with yield_self in Ruby 2.5 or then in Ruby 2.6
+        column_names.map { |column_name| all.map(&column_name.to_sym) }.tap { |values| break :zip.to_proc.(*values) }
```
kbrock, line 150: I think the bug you highlighted is we do not want to call `record[column_name]` but rather `record.public_send(column_name)`. (or `send`?)

This code has gone through at least 3 iterations:

```ruby
# first
column_names.map { |column_name| all.map(&column_name.to_sym) }.inject(&:zip)

#269
all.map do |record|
  symbolized_column_names.map { |column_name| record[column_name] }
end

#299
column_names.map { |column_name| all.map(&column_name.to_sym) }.then { :zip.to_proc.(*_1) }
```

I'm finding the `zip` and the `tap`/`then` a little complicated to mentally parse. (both versions of the zip code). Not sure what the extra traversals buy us.

Does something like this work for you?
(does send/public_send fix your bug?)
```ruby
all.map { |record| column_names.map { |column_name| record.public_send(column_name) } }
```

kbrock: Maybe the zip buys us some performance or something? Please share. Though, the non-zip version seems to be simpler for me.

iberianpig: Thank you for your great feedback.
 I initially implemented pluck in a way that was closer to the original code.
However considering your feedback, proceeding as you've recommended seems better, especially from a performance standpoint.

Here are the results of the benchmark script:

```ruby
        Benchmark.bm 25 do |r|
          r.report "pluck(first)" do
            100000 times { CountryWithContinent.pluck_with_first(:id, :name, :continent) }
          end

          r.report "pluck(#269)" do
            100000 times { CountryWithContinent.pluck_with_269(:id, :name, :continent) }
          end

          r.report "pluck(#299)" do
            100000 times { CountryWithContinent.pluck(:id, :name, :continent) }
          end

          r.report "pluck(public_send)" do
            100000 times { CountryWithContinent.pluck_with_public_send(:id, :name, :continent) }
          end
        end
```

```
                                user     system      total        real
pluck(first)                2.755767   0.009342   2.765109 (  2.770909)
pluck(#269)                 0.962630   0.003945   0.966575 (  0.966671)
pluck(#299)                 2.175707   0.008026   2.183733 (  2.184002)
pluck(public_send)          0.764322   0.000021   0.764343 (  0.764447)

```

I'll move forward with the public_send approach. This seems like a step in the right direction. Thanks for your suggestion :smile:

kbrock: oh wow. thanks.
Yes, you were definitely following the original code's style. I appreciate that.
After looking at the numbers, I don't feel so bad wanting to go with what I consider a simpler style.

kbrock (approved):
This looks great.

Since the pre 269 code uses the method access, this looks like it shouldn't introduce any unintended side effects

iberianpig:
@kbrock When is the release likely to include the contents of this pull request? :smile:

PR #303, opened 2024-04-24
test: pin sqlite3 dependency to `< 2.0`

flavorjones:
Because Rails's sqlite adapter requires sqlite3 `~> 1.4` through v7.1.3.2, tests started failing once the new sqlite3 v2.0 was pulled in:

https://github.com/active-hash/active_hash/actions/runs/8813099625

This has been fixed in Rails `7-1-stable` and `main` but no releases are yet available.

This PR:

- pins the test dependency to `< 2.0` for the relevant Rails versions
- and also sets `fail-fast: false` in the github actions pipeline matrix

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -20,6 +20,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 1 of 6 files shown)


kbrock (approved):
Lots of gems are also doing this

Issue #305, opened 2024-04-29
"`has_many': wrong number of arguments (given 2, expected 1) (ArgumentError)" error is raised since 3.3.0


issei-m:
[This PR](https://github.com/active-hash/active_hash/pull/296) introduces a new extension method `has_many` that overrides the one from `ActiveRecord::Base`, but it has a different argument signature, so the following code raises an error:

```ruby
class Country < ActiveHash::Base
end

class Post < ActiveRecord::Base
  belongs_to :person
end

class Person < ActiveRecord::Base
  extend ActiveHash::Associations::ActiveRecordExtensions
  belongs_to_active_hash :country

  # This association is for two ActiveRecord models
  has_many :posts, -> { order(:created_at) }, dependent: :destroy
end
```

This `has_many` takes three arguments according to the original ActiveRecord's method signature:

https://github.com/rails/rails/blob/v7.1.3.2/activerecord/lib/active_record/associations.rb#L1522

but extended one takes only two:

https://github.com/active-hash/active_hash/blob/v3.3.0/lib/associations/associations.rb#L9

Any idea to solve it? I'm using Rails 7.1.3.

kbrock:
@flavorjones ugh. we may need to check active record version when calling `super`

sontixyou:
I have created a PR that may solve this issue.
https://github.com/active-hash/active_hash/pull/306

flavorjones:
Closed by #306

flavorjones:
Fixed in https://github.com/active-hash/active_hash/releases/tag/v3.3.1

PR #306, opened 2024-04-30
Fix arguments of has_many method

sontixyou:
## Problem being solved

fix this issue. https://github.com/active-hash/active_hash/issues/305

I think that merging this PR will eliminate the error that is occurring in the above issue.

## Investigation

ActiveHash supports up to Rails ver5.0.
I checked Rails ver5.0 to Rails ver7.1
Then I found that the argument structure of the has_many method is almost the same.

- [Where has_many is defined in Rails v7.0](https://github.com/rails/rails/blob/v7.0.8.1/activerecord/lib/active_record/associations.rb#L1469)
- [Where has_many is defined in Rails v6.1](https://github.com/rails/rails/blob/v6.1.7.7/activerecord/lib/active_record/associations.rb#L1457)
- [Where has_many is defined in Rails v6.0](https://github.com/rails/rails/blob/v6.0.6.1/activerecord/lib/active_record/associations.rb#L1370)
- [Where has_many is defined in Rails v5.2](https://github.com/rails/rails/blob/v5.2.8.1/activerecord/lib/active_record/associations.rb#L1368)
- [Where has_many is defined in Rails v5.1](https://github.com/rails/rails/blob/v5.1.7/activerecord/lib/active_record/associations.rb#L1401)
- [Where has_many is defined in Rails v5.0](https://github.com/rails/rails/blob/v5.0.7.2/activerecord/lib/active_record/associations.rb#L1370)

Hence, I modified it to pass the same arguments as ActiveRecord#has_many.

### DIFF
--- a/lib/associations/associations.rb
+++ b/lib/associations/associations.rb
@@ -6,7 +6,7 @@ module ActiveHash
         require_relative 'reflection_extensions'
       end
 
-      def has_many(association_id, **options)
+      def has_many(association_id, scope = nil, **options, &extension)
         if options[:through]
           klass_name = association_id.to_s.classify
           klass =
--- a/spec/associations/active_record_extensions_spec.rb
+++ b/spec/associations/active_record_extensions_spec.rb
@@ -31,6 +31,32 @@ unless SKIP_ACTIVE_RECORD
       end
     end
 
+    def define_person_classes
+      define_ephemeral_class(:Country, ActiveHash::Base) do
+        self.data = [
+          {:id => 1, :name => "Japan"}
+        ]
+      end
+
+      define_ephemeral_class(:Person, ActiveRecord::Base) do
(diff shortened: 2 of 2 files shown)


flavorjones:
@sontixyou Thank you for submitting this. I would like to see a test for this usage, perhaps we can take the example @issei-m providede in #305?

sontixyou:
@flavorjones 
Thanks for the reply!
I have added a test on the changes.
Could you please review them?

issei-m (approved):
Once applied this patch, my issue described in #305 was resolved!

flavorjones (approved):

flavorjones:
@kbrock We should probably ship this in a bugfix release. You OK if I go ahead and do that?

flavorjones:
Shipped in https://github.com/active-hash/active_hash/releases/tag/v3.3.1

PR #309, opened 2024-05-06
fix: do not suppress errors when loading classes

andreynering:
This `rescue` block was suppressing real errors on our codebase. I had to remove this block on a local fork to be able to see the real cause of a bug that was silently happening.

I noticed that removing the blocks do not break any tests. Looks like we should just remove them?

### DIFF
--- a/lib/associations/associations.rb
+++ b/lib/associations/associations.rb
@@ -9,12 +9,7 @@ module ActiveHash
       def has_many(association_id, scope = nil, **options, &extension)
         if options[:through]
           klass_name = association_id.to_s.classify
-          klass =
-            begin
-              klass_name.safe_constantize
-            rescue StandardError, LoadError
-              nil
-            end
+          klass = klass_name.safe_constantize
 
           if klass && klass < ActiveHash::Base
             define_method(association_id) do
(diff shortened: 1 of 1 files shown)


kbrock:
@andreynering Thanks. good catch. I think it was missed when we converted the `constantize` to `safe_constantize` in #297

Can you share the example of what error was being squashed here? (for my own personal knowledge)

I did have concerns with the `constantize` here since loading the class here has caused race conditions in another codebase. But the added benefits of checking this class outweighed the rare issue.

Do you think you could put together a test here for when we specify an invalid class? (if it is super hard, then just say so and we can punt.

---

@flavorjones I missed this try/catch when you added the `safe_constantize`.
I'm assuming that we just missed it and I'd like to merge this change.

But I do have to ask: Did you leave this `try`/`catch` for a particular reason?
There is an issue when the inflection for the class is slightly different. (e.g.: class named `VM` but we try and load `Vm`) Is this the reason you left it in? Or was it just an oversight? This is a very rare edge case, but wanted to double check.

andreynering:
@kbrock We had a concern included on a given ActiveRecord model, but the content of `included do; end` block wasn't actually been ran. Ends up there was an error being raised, but that block was suppressing it.

flavorjones (approved):
I'm OK with it!

PR #311, opened 2024-07-03
Add column names method

### LINKED ISSUE #310: Add column_names Method to ActiveHash
## Description
I would like to propose adding a column_names method to ActiveHash, similar to the method available in ActiveRecord. The primary motivation for this addition is to facilitate the generation of CSV headers when exporting data from models that utilize ActiveHash::Base.

In ActiveRecord, the column_names method returns an array of column names as strings. The implementation is straightforward:

```ruby
# Returns an array of column names as strings.
def column_names
  @column_names ||= columns.map(&:name).freeze
end
```

## motivation

For ActiveHash, having a column_names method would be beneficial in scenarios where we need to export data to CSV, ensuring the headers are generated dynamically based on the model's columns.

## Use Case:

Exporting data from ActiveHash::Base models to CSV.
Dynamically generating CSV headers based on the model's columns.

hatsu38:
This pull request adds the column_names method to the ActiveHash gem. 

The column_names method is analogous to the one found in Rails' ActiveRecord, allowing users to retrieve the names of columns defined in an ActiveHash model. 

Added column_names method to ActiveHash::Base class.
The method returns an array of string column names, similar to ActiveRecord's column_names method.

```ruby
Country.column_names
# => ["id", "name", "code"]
```

close #310

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -51,7 +51,7 @@ module ActiveHash
       end
 
       def column_names
-        @column_names ||= field_names.map(&:to_s)
+        @column_names ||= field_names.map(&:name)
       end
 
       def the_meta_class
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -64,7 +64,9 @@ module ActiveHash
       #  # => ["id", "name", "code"]
       #
       def column_names
-        field_names.map(&:name)
+        field_names.map do |field|
+          field.respond_to?(:name) ? field.name : field.to_s.freeze
+        end
       end
 
       def the_meta_class
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -64,9 +64,7 @@ module ActiveHash
       #  # => ["id", "name", "code"]
       #
       def column_names
-        field_names.map do |field|
-          field.respond_to?(:name) ? field.name : field.to_s.freeze
-        end
+        field_names.map(&:name)
       end
 
       def the_meta_class
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -50,6 +50,10 @@ module ActiveHash
         @field_names ||= []
       end
 
+      def column_names
+        @column_names ||= field_names.map(&:to_s)
+      end
+
       def the_meta_class
         class << self
           self
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -64,7 +64,7 @@ module ActiveHash
       #  # => ["id", "name", "code"]
       #
       def column_names
-        @column_names ||= field_names.map(&:name)
+        field_names.map(&:name)
       end
 
       def the_meta_class
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -50,6 +50,19 @@ module ActiveHash
         @field_names ||= []
       end
 
+      #
+      # Useful for CSV integration needing column names as strings.
+      #
+      # @return [Array<String>] An array of column names as strings.
+      #
+      # @example Usage
+      #  class Country < ActiveHash::Base
+      #    fields :name, :code
+      #  end
+      #
+      #  Country.column_names
+      #  # => ["id", "name", "code"]
+      #
       def column_names
         @column_names ||= field_names.map(&:name)
       end
--- a/spec/active_hash/base_spec.rb
+++ b/spec/active_hash/base_spec.rb
@@ -116,6 +116,16 @@ describe ActiveHash, "Base" do
     end
   end
 
+  describe ".column_names" do
+    before do
+      Country.fields :name, :iso_name, "size"
+    end
+
+    it "returns an array of column names" do
+      expect(Country.column_names).to eq(["name", "iso_name", "size"])
+    end
+  end
+
   describe ".data=" do
     before do
       class Region < ActiveHash::Base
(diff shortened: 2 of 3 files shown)


lib/active_hash/base.rb
```diff
@@ -50,6 +50,10 @@ def field_names
         @field_names ||= []
       end
 
+      def column_names
```
flavorjones, line 53: I imagine the point here is that we need strings and not symbols for whatever integration with need with the CSV gem? If so, can you please document that in an rdoc string above the method?

lib/active_hash/base.rb
```diff
@@ -50,6 +50,10 @@ def field_names
...
+        @column_names ||= field_names.map(&:to_s)
```
flavorjones, line 54:
```suggestion
        @column_names ||= field_names.map(&:name)
```


Please use `Symbol.name` to avoid creating a new string for every invocation.

```
irb(main):005> :foo.to_s.frozen?
=> false
irb(main):006> :foo.to_s.object_id
=> 63900
irb(main):007> :foo.to_s.object_id
=> 65360
irb(main):008> :foo.to_s.object_id
=> 66820
```

vs

```
irb(main):016> :foo.name.frozen?
=> true
irb(main):017> :foo.name.object_id
=> 75280
irb(main):018> :foo.name.object_id
=> 75280
irb(main):019> :foo.name.object_id
=> 75280
```

flavorjones (changes requested):
@hatsu38 Thank you for submitting this!

If the goal here is to make sure active hash integrates with the CSV gem, can you please add a new spec that verifies this is working as expected? I'd suggest creating a new directory, `spec/integration`, and a new spec file.

I would also be interested in a short "how to generate CSV" in the README.md file.

flavorjones:
I would also like to see a short explicit test in `spec/active_hash/base_spec.rb` like the one I added in #312 for `#field_names`. Thank you!

kbrock:
Rails does not have `field_names`, instead they have `column_names`. So there is a difference.
Rails explicitly states that `column_names` is an array of strings.

I'd feel more comfortable with just doing the work for column_names.
@flavorjones what is your take?

```ruby
def column_names
  field_names.map(&:name)
end
```

Also looks like `field_names << field` is not ensuring that a symbol goes in there, but comparison routines assume it is a symbol. I'll create a PR for that.

lib/active_hash/base.rb
hatsu38: Got it! I didn't know that. I fix it ⚒️
https://github.com/active-hash/active_hash/pull/311/commits/0540d0301c9129ceae7fa56af1e8d2696ee44f4b

flavorjones:
@kbrock I agree, let's not cache column names, and always iterate over field names when it's called. @hatsu38 Can you make that change?

lib/active_hash/base.rb
hatsu38: Ruby versions prior to 3.0 do not have the name method in the Symbol class. CI Fail [Job].(https://github.com/active-hash/active_hash/actions/runs/9796570215/job/27117706794) 

So, I fix it again https://github.com/active-hash/active_hash/pull/311/commits/2f6b35d4a0c5f21ec5365bb36668cb12be5b7551

lib/active_hash/base.rb
```diff
@@ -50,6 +50,25 @@ def field_names
...
+      #    fields :name, :code
+      #  end
+      #
+      #  Country.column_names
+      #  # => ["id", "name", "code"]
+      #
+      def column_names
+        field_names.map do |field|
+          field.respond_to?(:name) ? field.name : field.to_s.freeze
+        end
```
kbrock, line 69: Now field names are all symbols, so I think we can simplify this:

```suggestion
        fields.map(&:name)
```

Does that work for you?

hatsu38: After merging the latest master, the code worked perfectly! https://github.com/active-hash/active_hash/pull/311/commits/030685854a7876130595b630c283fbba7265660e

kbrock (approved):
thanks

kbrock:
@flavorjones Mike, `Symbol.name` was introduced in ruby 3.0. Do you have an opinion on how to resolve?

I almost think dropping Ruby <3.0 (and therefore rails rails <=6.0. It sounds drastic but they are all EOL for a while and this will work for them, just not the new `column_names`.

OTOH/ we could use `Symbol.to_s` and not worry about the frozen world.

flavorjones:
Oof, hmm. Well for the record: Rails 6.1 is still supported, and it still technically supports Ruby 2.7. Totally annoying.

But: you're right that this is a new feature, and Rails 6.1 is only supported for security updates at this point. So I agree it is feasible and probably preferable to drop support for Ruby < 3.0 in the next minor release.

spec/active_hash/base_spec.rb
```diff
@@ -116,6 +116,16 @@ class Country < ActiveHash::Base
...
   end
 
+  describe ".column_names" do
+    before do
+      Country.fields :name, :iso_name, "size"
+    end
+
+    it "returns an array of column names" do
+      expect(Country.column_names).to eq(["name", "iso_name", "size"])
+    end
```
kbrock, line 126: @flavorjones do we want to just conditionally do this?

```suggestion
    if RUBY_VERSION >= "3.0"
    it "returns an array of column names" do
      expect(Country.column_names).to eq(["name", "iso_name", "size"])
    end
    end
```

flavorjones: Seems fine to me.

hatsu38: I commited! https://github.com/active-hash/active_hash/pull/311/commits/9679a8998128c2e11c994e76dde132135759f153

kbrock (approved):
Hard fought. This looks good.

@flavorjones The githubs humbly requests you take a look at this too.

flavorjones (approved):

PR #312, opened 2024-07-03
fix: ensure the stored field names are always symbols

flavorjones:
There is code that assumes that `@field_names` will contain symbols, like this method in ActiveHash::Base:

    def respond_to?(method_name, include_private=false)
      super ||
        begin
          config = configuration_for_custom_finder(method_name)
          config && config[:fields].all? do |field|
            field_names.include?(field.to_sym) || field.to_sym == :id
          end
        end
    end

This also introduces an explicit test for `#field_names`.

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -198,7 +198,9 @@ module ActiveHash
       end
 
       def field(field_name, options = {})
+        field_name = field_name.to_sym
         validate_field(field_name)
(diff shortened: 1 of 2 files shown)


kbrock (approved):
Yours makes fewer changes.

I'm a little amused that I asked if we could use strings for field/column names, yet we both doubled down on using symbols for column names.

PR #313, opened 2024-07-03
Ensure field_names contains only symbols

kbrock:
All public methods that take a field name will convert it to a symbol.
All internal methods can assume that field_name is a symbol

In response to a bug that was uncovered in #311 (test is red-green)

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -198,6 +198,7 @@ module ActiveHash
       end
 
       def field(field_name, options = {})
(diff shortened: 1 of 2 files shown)


flavorjones (approved):
I think I beat you by a few hours with #312 but I don't really care which one we merge. You pick!

kbrock:
Thanks for the comment. Didn't see that one (heh - obviously or I wouldn't have created this one)

kbrock:
defer to #312

PR #314, opened 2024-07-08
Make Default field private

kbrock:
Followup to #312 

I noticed that this is public.

We either want to make this `private
Or we want to `to_sym` the field name.

I went with the first, but just say if you would rather make it `public` and I can add a `to_sym` instead.

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -266,6 +266,8 @@ module ActiveHash
         self.default_attributes[field_name] = default_value
       end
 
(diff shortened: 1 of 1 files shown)


flavorjones (approved):
LGTM, I think we should make this private, people should be using the `field(default:)` option.

kbrock:
update:
- sorry. removed first commit which was unrelated

This is now just converting a private field

flavorjones:
LGTM!

PR #317, opened 2024-07-25
Add support for a block argument on `Relation#count`

### LINKED ISSUE #316: `count` should support a block
### Description

Hi! Thanks for this gem - it has been a go-to for readonly data in our projects.

We ran into an issue in our application where calling `count` with a block on a relation returned the count of elements in the relation, not the count of elements for which the block returns true.

Would it be possible for `count` to fallback to `Enumerable#count` if given a block? Rails introduced this in [5.1](https://github.com/rails/rails/pull/24203) for `ActiveRecord::Relation` and I think it'd made sense for `ActiveHash::Relation` to behave the same.

### Code to Reproduce
```ruby
class Example < ActiveHash::Base
  self.data = [
    { id: 1, active: true },
    { id: 2, active: true },
    { id: 3, active: false }
  ]
end

Example.count
# 3
Example.count { |e| e.active }
# 3
Example.all.to_a.count { |e| e.active }
# 2

# Desired behavior:
# Example.count { |e| e.active }
# 2
```

ashleyHutton:
Closes #316 

This pull request adds support for a block argument in the `#count` method.

### DIFF
--- a/lib/active_hash/relation.rb
+++ b/lib/active_hash/relation.rb
@@ -135,6 +135,7 @@ module ActiveHash
     end
 
     def count
+      return super if block_given?
       length
     end
 
--- a/spec/active_hash/relation_spec.rb
+++ b/spec/active_hash/relation_spec.rb
@@ -39,6 +39,16 @@ RSpec.describe ActiveHash::Relation do
     end
   end
 
+  describe '#count' do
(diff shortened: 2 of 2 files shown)


flavorjones (approved):
LGTM!

kbrock:
This looks good.

PR #321, opened 2024-08-27
Implement automatic predicate method generation for ActiveHash::Enum

### LINKED ISSUE #319: Add built-in predicate methods for enum types in ActiveHash
## Description
When using ActiveHash with enum types, we often find ourselves writing repetitive predicate methods for each enum value. For example:

```ruby
class PublicStatus < ActiveHash::Base
  include ActiveHash::Enum

  self.data = [
    { id: 1, name: "Publish", type: "published" },
    { id: 2, name: "Draft", type: "drafted" },
    { id: 3, name: "Archive", type: "archived" }
  ]

  enum_accessor :type

  def published?
    id == PublicStatus::PUBLISHED.id
  end

  def drafted?
    id == PublicStatus::DRAFTED.id
  end

  def archived?
    id == PublicStatus::ARCHIVED.id
  end
end
```

These methods are commonly used but require manual implementation for each enum value. It would be beneficial if ActiveHash could automatically generate these status check methods based on the defined enum values.

## Proposed Solution
Enhance ActiveHash to automatically generate status check methods for enum values. This could be implemented as follows:

When enum_accessor :type is called, automatically generate methods for each enum value.
The generated methods would follow the pattern {enum_value}? and return a boolean indicating whether the current object matches that enum value.

## Example Usage
With this enhancement, the above code could be simplified to:

```ruby
class PublicStatus < ActiveHash::Base
  include ActiveHash::Enum

  self.data = [
    { id: 1, name: "Publish", type: "published" },
    { id: 2, name: "Draft", type: "drafted" },
    { id: 3, name: "Archive", type: "archived" }
  ]

  enum_accessor :type
  # No need to manually define published?, drafted?, archived? methods
end

status = PublicStatus.find(1)
status.published? => true
status.drafted?   => false
status.archived?  => false
```

hatsu38:
close #319 

## Description
This PR addresses the issue of repetitive predicate method definitions in ActiveHash when using enum types. It implements an automatic generation of status check methods based on the defined enum values, simplifying the usage of ActiveHash::Enum.

## Changes
- Added `define_enum_methods` to `ActiveHash::Enum::Methods` module
- Automatically generate predicate methods when `enum_accessor` is called
- Sanitize method names by removing non-word characters and leading/trailing underscores
- Define instance methods that compare record IDs for enum value checking

## Example
Before:
```ruby
class PublicStatus < ActiveHash::Base
  include ActiveHash::Enum
  self.data = [
    { id: 1, name: "Publish", type: "published" },
    { id: 2, name: "Draft", type: "drafted" },
    { id: 3, name: "Archive", type: "archived" }
  ]
  enum_accessor :type

  def published?
    id == PublicStatus::PUBLISHED.id
  end

  def drafted?
    id == PublicStatus::DRAFTED.id
  end

  def archived?
    id == PublicStatus::ARCHIVED.id
  end
end
```

After:
```ruby
class PublicStatus < ActiveHash::Base
  include ActiveHash::Enum
  self.data = [
    { id: 1, name: "Publish", type: "published" },
    { id: 2, name: "Draft", type: "drafted" },
    { id: 3, name: "Archive", type: "archived" }
  ]
  enum_accessor :type
  # No need to manually define published?, drafted?, archived? methods
end

status = PublicStatus.find(1)
status.published? # => true
status.drafted?   # => false
status.archived?  # => false
```

## Testing

Added unit tests to verify the automatic generation of predicate methods
Tested with various enum value names to ensure proper sanitization

### DIFF
--- a/lib/enum/enum.rb
+++ b/lib/enum/enum.rb
@@ -14,6 +14,17 @@ module ActiveHash
         reload
       end
 
+      def enum(columns)
+        columns.each do |column, values|
+          values = values.zip(values.map(&:to_s)).to_h if values.is_a?(Array)
+          values.each do |method, value|
+            define_method("#{method}?") do
+              send(column) == value
+            end
+          end
+        end
+      end
+
       def insert(record)
         super
         set_constant(record) if defined?(@enum_accessors)
--- a/lib/enum/enum.rb
+++ b/lib/enum/enum.rb
@@ -19,7 +19,7 @@ module ActiveHash
         columns.each do |column, values|
           values = values.zip(values.map(&:to_s)).to_h if values.is_a?(Array)
           values.each do |method, value|
-            method_definitions << "def #{method}?; #{column} == '#{value}'; end"
+            method_definitions << "def #{method}?; #{column} == #{value.inspect}; end"
           end
         end
         class_eval(method_definitions.uniq.join(";"))
--- a/lib/enum/enum.rb
+++ b/lib/enum/enum.rb
@@ -15,14 +15,17 @@ module ActiveHash
       end
 
       def enum(columns)
-        method_definitions = []
         columns.each do |column, values|
           values = values.zip(values.map(&:to_s)).to_h if values.is_a?(Array)
           values.each do |method, value|
-            method_definitions << "def #{method}?; #{column} == #{value.inspect}; end"
+            class_eval <<~METHOD, __FILE__, __LINE__ + 1
+              # frozen_string_literal: true
+              def #{method}?
+                #{column} == #{value.inspect}
+              end
+            METHOD
           end
         end
-        class_eval(method_definitions.uniq.join(";"))
       end
 
       def insert(record)
--- a/lib/enum/enum.rb
+++ b/lib/enum/enum.rb
@@ -15,14 +15,14 @@ module ActiveHash
       end
 
       def enum(columns)
+        method_definitions = []
         columns.each do |column, values|
           values = values.zip(values.map(&:to_s)).to_h if values.is_a?(Array)
           values.each do |method, value|
-            define_method("#{method}?") do
-              send(column) == value
-            end
+            method_definitions << "def #{method}?; #{column} == '#{value}'; end"
           end
         end
+        class_eval(method_definitions.uniq.join(";"))
       end
 
       def insert(record)
--- a/spec/enum/enum_spec.rb
+++ b/spec/enum/enum_spec.rb
@@ -79,6 +79,73 @@ describe ActiveHash::Base, "enum" do
       expect(Movie::THE_INFORMANT.name).to eq('The Informant!')
       expect(Movie::IN_OUT.name).to eq('In & Out')
     end
+
+    describe "enum(columns)" do
+      it "defines a predicate method for each value in the enum" do
+        Article = Class.new(ActiveHash::Base) do
+          include ActiveHash::Enum
+
+          self.data = [
+            { name: 'Article 1', status: 'draft'},
+            { name: 'Article 2', status: 'published'},
+            { name: 'Article 3', status: 'archived'}
+          ]
+
+          enum_accessor :name
+
+          enum status: [:draft, :published, :archived]
+        end
+
+        expect(Article::ARTICLE_1.draft?).to be_truthy
+        expect(Article::ARTICLE_1.published?).to be_falsey
+        expect(Article::ARTICLE_1.archived?).to be_falsey
+
+        expect(Article::ARTICLE_2.draft?).to be_falsey
+        expect(Article::ARTICLE_2.published?).to be_truthy
+        expect(Article::ARTICLE_2.archived?).to be_falsey
+
+        expect(Article::ARTICLE_3.draft?).to be_falsey
+        expect(Article::ARTICLE_3.published?).to be_falsey
+        expect(Article::ARTICLE_3.archived?).to be_truthy
+      end
+
+      it "defines a predicate method for each value in the enum" do
+        NotifyType = Class.new(ActiveHash::Base) do
+          include ActiveHash::Enum
+
+          self.data = [
+            { name: 'Like', action: 'LIKE'},
+            { name: 'Comment', action: 'COMMENT'},
+            { name: 'Follow', action: 'FOLLOW'},
+            { name: 'Mention', action: 'MENTION'}
+          ]
+
+          enum_accessor :name
+
+          enum action: { like: 'LIKE', comment: 'COMMENT', follow: 'FOLLOW', mention: 'MENTION' }
+        end
+
+        expect(NotifyType::LIKE.like?).to be_truthy
+        expect(NotifyType::LIKE.comment?).to be_falsey
+        expect(NotifyType::LIKE.follow?).to be_falsey
+        expect(NotifyType::LIKE.mention?).to be_falsey
+
+        expect(NotifyType::COMMENT.like?).to be_falsey
+        expect(NotifyType::COMMENT.comment?).to be_truthy
+        expect(NotifyType::COMMENT.follow?).to be_falsey
+        expect(NotifyType::COMMENT.mention?).to be_falsey
+
+        expect(NotifyType::FOLLOW.like?).to be_falsey
+        expect(NotifyType::FOLLOW.comment?).to be_falsey
+        expect(NotifyType::FOLLOW.follow?).to be_truthy
+        expect(NotifyType::FOLLOW.mention?).to be_falsey
+
+        expect(NotifyType::MENTION.like?).to be_falsey
+        expect(NotifyType::MENTION.comment?).to be_falsey
+        expect(NotifyType::MENTION.follow?).to be_falsey
+        expect(NotifyType::MENTION.mention?).to be_truthy
+      end
+    end
   end
 
   context "ActiveHash with an enum_accessor set" do
--- a/spec/enum/enum_spec.rb
+++ b/spec/enum/enum_spec.rb
@@ -109,20 +109,20 @@ describe ActiveHash::Base, "enum" do
         expect(Article::ARTICLE_3.archived?).to be_truthy
       end
 
-      it "defines a predicate method for each value in the enum" do
+      it "multi type data (ex: string, integer and symbol) enum" do
         NotifyType = Class.new(ActiveHash::Base) do
           include ActiveHash::Enum
 
           self.data = [
             { name: 'Like', action: 'LIKE'},
-            { name: 'Comment', action: 'COMMENT'},
-            { name: 'Follow', action: 'FOLLOW'},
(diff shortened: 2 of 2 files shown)


lib/enum/enum.rb
```diff
@@ -29,6 +28,20 @@ def delete_all
         super
       end
 
+      private
+
+      def define_enum_methods
+        return if @enum_accessors.blank?
+
+        @records&.each do |record|
+          enum_vallue = @enum_accessors.map { |name| record.attributes[name] }.join("_").downcase.gsub(/\W+/, '_').gsub(/^_|_$/, '')
```
flavorjones, line 37: Please name this variable `enum_value` which is the conventional spelling.

lib/enum/enum.rb
```diff
@@ -40,8 +53,6 @@ def set_constant(record)
         end
       end
 
-      private :set_constant
```
flavorjones, line 43: Just a note that this change is fine and the methods in question are still private because of the `private` declaration above. That said, I would very much prefer in the future that changes like this are presented in a separate refactoring commit.

lib/enum/enum.rb
flavorjones, line 37: Did this subexpression come from somewhere else? Why not use something like Active Support's `#camelize` instead of this?

Also, how is this implementation going to handle collisions (e.g., two values like "FooBar" and "foo_bar" that would create the same method name)? How does rails handle collisions? I would like to see some test coverage for a collision case as well

flavorjones (changes requested):
I think we need to make this opt-in behavior (and not default behavior) because people may have already implemented their own methods that, depending on when the Enum module is included, might be clobbered during an upgrade.

I'd love to hear your thoughts on how to handle that case.

lib/enum/enum.rb
hatsu38: Fix: https://github.com/active-hash/active_hash/pull/321/commits/9e2fc62b37e5f108ed06ba2e7cd25fed40df7fb5

lib/enum/enum.rb
hatsu38: Fix: https://github.com/active-hash/active_hash/pull/321/commits/9e2fc62b37e5f108ed06ba2e7cd25fed40df7fb5

hatsu38: The enum_value was derived from this section.
https://github.com/active-hash/active_hash/blob/a192385350b8abc11b77937e3556ad9f3fde587e/lib/enum/enum.rb#L48-L49

> Also, how is this implementation going to handle collisions (e.g., two values like "FooBar" and "foo_bar" that would create the same method name)?

I try it. 
A DuplicateEnumAccessor error occurs when the set_constant method is executed.

https://github.com/active-hash/active_hash/blob/a192385350b8abc11b77937e3556ad9f3fde587e/lib/enum/enum.rb#L39

And We have also added a test for cases where the method is already defined by itself.
https://github.com/active-hash/active_hash/pull/321/commits/00c149770bedfb7d5798dc2c44b65349ca8c93c6

flavorjones:
@kbrock wdyt of this feature

kbrock:
@flavorjones I had started to throw some code together to see how I would implement it. I'm glad this PR came through before I got too far.

Not really a fan of this, but I appreciate that we all have our own preferred styles.

https://api.rubyonrails.org/v5.1/classes/ActiveRecord/Enum.html

Docs have:

```ruby
class Conversation < ActiveRecord::Base
  enum status: [ :active, :archived ]
end

conversation.active? # => true
conversation.status  # => "active"
```

A quick implementation would probably be like:
(think they have some code cache type framework to make this a little easier)

```ruby
def enum columns
  methods = []
  columns.each do |column, values|
    values.each do |value|
      methods << "def #{value}? ; #{column} == #{value.inspect} ; end"
    end
  end
  instance_eval { methods.uniq.join(";") }
end
```

So then that begs to ask, why do we define a constant for each class in enum? This feels very different from active record.
This new request is more similar, though I'd just compare the column value to the constant rather than dynamically adding values for each record.

I want to accept this, but feel my negativity coming through. I do not understand why I'm negative though. It seems relatively sound.

Tangent: the while undefining a constant when the field is removed distracts me. Don't think we should be supporting delete/insert/update type behavior. again, not related though

hatsu38:
Thank you for the insightful comment!

> Don't think we should be supporting delete/insert/update type behavior. again, not related though

I agree with this.

I personally prefer the following style, where when calling `enum_accessor`, methods like `drafted?` are also available. This is because I often use `enum_accessor`.

```ruby
class PublicStatus < ActiveHash::Base
  include ActiveHash::Enum
  self.data = [
    { id: 1, name: "Publish", type: "published" },
    { id: 2, name: "Draft", type: "drafted" },
    { id: 3, name: "Archive", type: "archived" }
  ]
  enum_accessor :type
  # No need to manually define published?, drafted?, archived? methods
end
```

Do you prefer the approach of not using `enum_accessor`, like the one below?

```ruby
コードをコピーする
class PublicStatus < ActiveHash::Base
  include ActiveHash::Enum
  self.data = [
    { id: 1, name: "Publish", type: "published" },
    { id: 2, name: "Draft", type: "drafted" },
    { id: 3, name: "Archive", type: "archived" }
  ]

  enum type: [:published, :drafted, :archived]
end
```

I'm currently debating the implementation approach.

lib/enum/enum.rb
```diff
@@ -29,6 +28,20 @@ def delete_all
...
       end
 
+      def define_enum_methods
+        return if @enum_accessors.blank?
+
+        @records&.each do |record|
+          enum_value = @enum_accessors.map { |name| record.attributes[name] }.join("_").downcase.gsub(/\W+/, '_').gsub(/^_|_$/, '')
+          method_name = "#{enum_value}?"
+          define_method(method_name) do
+            self.id == record.id
```
kbrock, line 38: Yea. Instead of Id, just compare the value. This will allow duplicates and work closer to rails.

kbrock:
@hatsu38 Thank you for your logical thoughts.

```diff
-  enum_accessor :type
+  enum type: [:published, :drafted, :archived]
```

- I like the first because you have less duplication.
- I like the second because you can easily include that line before and after the data is set.
- I like the second because it looks more similar to active record.

The current implementation to add constants is a little troubling. I don't like that it keeps adding constants to the class for every record added. I would like to avoid adding different methods to the class for every record added. I'd prefer to make as many changes all at once.

If we agree that `enum` must be defined after `data`, then `enum :type, :type2` also seems acceptable. Not sure if documentation will help people avoid mistakes or if it is inviting support issues.

hatsu38:
@kbrock Thanks you!

The implementation with

```ruby
enum type: [:published, :drafted, :archived]
```

is closer to Rails, making it easier to understand for those who are new to ActiveHash. After actually trying it out, it feels pretty good to write! I think I'll proceed with this approach 🙆

kbrock:
Also, though my opinion has changed over the years, I think I'm in camp `instance_eval {}` vs `def_method()`. All those closures hanging about makes me nervous.

hatsu38:
> Also, though my opinion has changed over the years, I think I'm in camp `instance_eval {}` vs `def_method()`. All those closures hanging about makes me nervous.

I wasn't sure how to use instance_eval, so I tried using class_eval instead.
If you notice anything that could be improved, I'd appreciate your feedback in the review! 👍

lib/enum/enum.rb
```diff
@@ -14,6 +14,17 @@ def enum_accessor(*field_names)
         reload
       end
 
+      def enum(columns)
+        method_definitions = []
+        columns.each do |column, values|
+          values = values.zip(values.map(&:to_s)).to_h if values.is_a?(Array)
+          values.each do |method, value|
+            method_definitions << "def #{method}?; #{column} == '#{value}'; end"
```
kbrock, line 22:
```suggestion
            method_definitions << "def #{method}?; #{column} == #{value.inspect}; end"
```

Will `value.inspect` work?
I'd prefer to use that if it buys us enumeration with columns that contain integers.

Strings are most important, integers are a nice to have, and symbols are a very distant nice to have. I think we get all 3 with inspect

spec/enum/enum_spec.rb
```diff
@@ -79,6 +79,73 @@ class Neighborhood < ActiveHash::Base
...
+          self.data = [
+            { name: 'Like', action: 'LIKE'},
+            { name: 'Comment', action: 'COMMENT'},
+            { name: 'Follow', action: 'FOLLOW'},
+            { name: 'Mention', action: 'MENTION'}
+          ]
+
+          enum_accessor :name
+
+          enum action: { like: 'LIKE', comment: 'COMMENT', follow: 'FOLLOW', mention: 'MENTION' }
```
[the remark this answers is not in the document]
kbrock: good find. this looks great

lib/enum/enum.rb
hatsu38: I have decided to go with your suggestion!
https://github.com/active-hash/active_hash/pull/321/commits/5cab8ac5e9d83af11093997e1cb24955a1a999ad

```ruby
bundle exec rspec spec/enum/enum_spec.rb   
...........

Finished in 0.01806 seconds (files took 0.49792 seconds to load)
11 examples, 0 failures
```

kbrock (approved):

kbrock:
This looks great.

I tried do an injection attack here and was not able to get anything through.
(I was able to sneak in one for `'#{value}'`)

The closes I can come is:

```ruby
x="a'||true||'"
instance_eval("def a; 'z' == '#{x}' ; end") ; a
# => true

x=:"a||true"
instance_eval("def a; 'z' == #{x.inspect} ; end") ; a
# => false
```

@flavorjones you good to go on this one?

flavorjones (approved):
I've pushed a commit that sets the line number for `eval` and uses frozen strings -- these are Rails project conventions that I think are good.

I've also rebased onto master.

flavorjones:
Whoops, force-pushed a fix to the frozen strings comment.

lib/enum/enum.rb
```diff
@@ -14,6 +14,20 @@ def enum_accessor(*field_names)
...
+      def enum(columns)
+        columns.each do |column, values|
+          values = values.zip(values.map(&:to_s)).to_h if values.is_a?(Array)
+          values.each do |method, value|
+            class_eval <<~METHOD, __FILE__, __LINE__ + 1
+              # frozen_string_literal: true
+              def #{method}?
+                #{column} == #{value.inspect}
+              end
+            METHOD
```
kbrock, line 26: Great call on `fronzen_string_literal` and the line number.

I had taken my inspiration from https://github.com/rails/rails/blob/main/activesupport/lib/active_support/core_ext/class/attribute.rb#L119-L120

They only call `class_eval` once, hence the `method_definitions`, and they use the caller location. But this should work fine.

Interestingly, enum uses `def_method`.

^ more of a note to future self.

PR #326, opened 2024-12-17
Fix an example for resetting the state every request

y-yagi:
`before_filter` was removed by Rails v5.1.0.


flavorjones (approved):
Thank you!

PR #329, opened 2025-06-03
Has many through supports aliased names

alexgriff:
### Bug/Problem

Currently, using a `has_many :through` to return a collection of ActiveHash records is limited to working successfully in the case where the `association_id` passed to `has_many` exactly corresponds to the ActiveHash class name. It must be the pluralized version of the singularized class name.

In this example, `Zoo#countries` works as expected:
```rb
  # ...
end

class Animal < ApplicationRecord
  
  belongs_to :country
  belongs_to :zoo
end

class Zoo < ApplicationRecord

  has_many :animals
  has_many :countries, through: :animals
end
```

But, If any of the association method names involved in the `has_many through:` require some type of aliasing where the relationship to the ActiveHash class can't be deduced, the behavior in `ActiveHash::Associations::ActiveRecordExtensions` will not correctly define the ActiveHash version of the association reader method.

Here, `Zoo#nations` would break
```rb
class Zoo < ApplicationRecord

  has_many :animals
  has_many :nations, through: :animals, source: :country
end
```

And here, `Zoo#countries` would break
```rb
class Animal < ApplicationRecord
  
  belongs_to :country_of_origin, class_name: 'Country'
  belongs_to :zoo
end

class Zoo < ApplicationRecord

  has_many :animals
  has_many :countries, through: :animals
end
```

It felt worth supporting these cases because it feels similar to how [the `class_name` option is correctly supported in a `belongs_to_active_hash`](https://github.com/active-hash/active_hash/blob/master/lib/associations/associations.rb#L43) for equivalent use cases around aliased association names.

I made sure the use of `source` matches how that option is passed [in ActiveRecord itself](https://apidock.com/rails/ActiveRecord/Associations/ClassMethods/has_many#:~:text=%3Asource,source%20is%20given.) (`has_many throughs:` use `source`, and don't support `class_name`).

### Documentation

I looked for a place to document this new behavior in the README, but there wasn't an obvious place to add this without adding a whole new section or something like that. I can try to do so if that feels appropriate.

### Specs
I added specs for the new supported cases. To reuse the same spec setup and ephemeral classes, but at the same time to be able to change the association definitions per spec context, I parameterized the `define_doctor_classes` spec setup method. This adds some complexity in a way I don't love, but still felt better than alternate approaches. I'm game to rethink this, though, if it doesn't feel like the right solution.

### DIFF
--- a/lib/associations/associations.rb
+++ b/lib/associations/associations.rb
@@ -8,16 +8,16 @@ module ActiveHash
 
       def has_many(association_id, scope = nil, **options, &extension)
         if options[:through]
-          active_hash_source_association_name = options[:source]&.to_s || association_id.to_s.singularize
+          source_association_name = options[:source]&.to_s || association_id.to_s.singularize
 
           through_klass = reflect_on_association(options[:through])&.klass
-          klass = through_klass&.reflect_on_association(active_hash_source_association_name)&.klass
+          klass = through_klass&.reflect_on_association(source_association_name)&.klass
 
           if klass && klass < ActiveHash::Base
             define_method(association_id) do
               join_models = send(options[:through])
               join_models.flat_map do |join_model|
-                join_model.send(active_hash_source_association_name)
+                join_model.send(source_association_name)
               end.uniq
             end
 
--- a/lib/associations/associations.rb
+++ b/lib/associations/associations.rb
@@ -8,14 +8,16 @@ module ActiveHash
 
       def has_many(association_id, scope = nil, **options, &extension)
         if options[:through]
-          klass_name = association_id.to_s.classify
-          klass = klass_name.safe_constantize
+          active_hash_source_association_name = options[:source]&.to_s || association_id.to_s.singularize
+
+          through_klass = reflect_on_association(options[:through])&.klass
+          klass = through_klass&.reflect_on_association(active_hash_source_association_name)&.klass
 
           if klass && klass < ActiveHash::Base
             define_method(association_id) do
               join_models = send(options[:through])
               join_models.flat_map do |join_model|
-                join_model.send(association_id.to_s.singularize)
+                join_model.send(active_hash_source_association_name)
               end.uniq
             end
 
--- a/spec/associations/active_record_extensions_spec.rb
+++ b/spec/associations/active_record_extensions_spec.rb
@@ -85,7 +85,12 @@ unless SKIP_ACTIVE_RECORD
       define_ephemeral_class(:SchoolStatus, ActiveHash::Base)
     end
 
-    def define_doctor_classes
+    def define_doctor_classes(
+      appointment_to_physician_association_name:,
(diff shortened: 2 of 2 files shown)


kbrock (commented):
this looks like a nice change.

1. potential shortening of has_many local variable name
2. simplification of tests
3. squashing the commits

(though you may want to squash commits, make my suggested change in a separate commit - in case we decide it doesn't make it any better)

spec/associations/active_record_extensions_spec.rb
```diff
@@ -234,6 +234,33 @@ def define_doctor_classes
 
           expect(patient.physicians).to contain_exactly(physician1, physician2)
         end
+
+        describe "with the :source option" do
+          it "finds ActiveHash records through the join model" do
+            Patient.has_many :doctors, through: :appointments, source: :physician
```
alexgriff, line 240: @kbrock what do you think of this approach where we simply add the differently named associations and dont bother removing the "default" ones. I was kind of wanting to not have both associations there in the spec contexts, but it's not quite necessary and doesn't really effect the functionality.

One other thought is that I could take the approach here but also explicitly remove the associations and added methods, something like:
```rb
Appointment._reflections.delete('physician')
Appointment.class_eval do
  undef_method :physician
  undef_method :physician=
end
Appointment.belongs_to :doctor, class_name: 'Physician', foreign_key: :physician_id
```
... but that seems like a lot for something that really doesn't "do anything"; specs will pass with or without it.

lib/associations/associations.rb
```diff
@@ -8,14 +8,16 @@ def self.extended(base)
 
       def has_many(association_id, scope = nil, **options, &extension)
         if options[:through]
-          klass_name = association_id.to_s.classify
-          klass = klass_name.safe_constantize
+          source_association_name = options[:source]&.to_s || association_id.to_s.singularize
```
kbrock, line 11: not sure why, but I do feel this is much easier to read.
If you are having trouble with it, you can add back.

alexgriff:
@kbrock I think I'm all set - thanks for shepherding this through!

> I think I prefer less code, but totally up to you.

I get this, but I ended up definitely going "more code" not less, and actually deleting the old associations.

I liked that I could get an actual error messages in there easily enough which also kinda acts like a comment
```rb
define_method(:physicians) { raise NoMethodError, "The #physicians association is removed in this spec, use #doctors" }
```
But happy to change anything

spec/associations/active_record_extensions_spec.rb
```diff
@@ -234,6 +234,56 @@ def define_doctor_classes
 
           expect(patient.physicians).to contain_exactly(physician1, physician2)
         end
+
+        describe "with the :source option" do
+          before do
+            # NOTE: Removing the Patient#physicians association and adding Patient#doctors
```
kbrock, line 240: perfect

PR #335, opened 2025-08-07
Fix `model_name` availability by explicitly extending ActiveModel::Naming

moznion:
In v4.0.0, switching from `extend ActiveModel::Naming` to only `extend ActiveModel::Translation` (ref: https://github.com/active-hash/active_hash/pull/230) can cause "undefined method 'model_name'" errors when ActiveHash models include `ActiveModel::Serialization`, particularly with Rails 8.0+; actually I encountered this problem in our project.

While `ActiveModel::Translation` includes `ActiveModel::Naming` internally, the inclusion doesn't always properly expose the `model_name` class method when extended. This breaks compatibility for models that depend on `model_name` being available at the class level.

This commit explicitly extends both `ActiveModel::Naming` and `ActiveModel::Translation` to ensure model_name is always available while maintaining i18n support. This approach ensures compatibility across all Rails versions.

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -24,6 +24,7 @@ module ActiveHash
     class_attribute :_data, :dirty, :default_attributes, :scopes
 
     if Object.const_defined?(:ActiveModel)
+      extend ActiveModel::Naming
       extend ActiveModel::Translation
       include ActiveModel::Conversion
     else


kbrock:
ok, so this is a swap. Sounds like a straight forward solution.

I also saw one PR asking to add JSON, and worried that we will unnecessarily include the kitchen sink.

Helping us come up with a test case would be great.

moznion:
@kbrock @flavorjones Thank you for your suggestions.
I just added a test case in [9c4af39](https://github.com/active-hash/active_hash/pull/335/commits/9c4af3964012c51f8dae936542c3b184ed91deea). This test is quite simple but you can reproduce the problem by commenting out `extend ActiveModel::Naming` in `lib/active_hash/base.rb`.

PR #336, opened 2025-08-18
Fix deadlock in belongs_to/has_many :through; support polymorphic source_type

### LINKED ISSUE #334: Polymorphic `has_many` with a `source_type` raise an exception in ActiveHash 4.0
The following association:

```rb
has_many :asset_types, through: :assignments, source: :assignable, source_type: "AssetType"
```

raises the following exception:

```rb
/Users/jpcody/.local/share/mise/installs/ruby/3.3.8/lib/ruby/gems/3.3.0/gems/activerecord-7.1.5.1/lib/active_record/reflection.rb:464:in `compute_class': Polymorphic associations do not support computing the class. (ArgumentError)

          raise ArgumentError, "Polymorphic associations do not support computing the class."
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        from /Users/jpcody/.local/share/mise/installs/ruby/3.3.8/lib/ruby/gems/3.3.0/gems/activerecord-7.1.5.1/lib/active_record/reflection.rb:412:in `klass'
        from /Users/jpcody/.local/share/mise/installs/ruby/3.3.8/lib/ruby/gems/3.3.0/gems/active_hash-4.0.0/lib/associations/associations.rb:14:in `has_many'
        from /Users/jpcody/Projects/CampusOptics/Code/optics-web/app/models/user.rb:38:in `<class:User>'
```

In our codebase. We follow this association pattern in multiple places, and while I tried to dive in a bit, we're also on the verge of upgrading Rails. For now, I'm going to monkeypatch AH to the old version and see if a newer version of Rails resolves this. But I wanted to go ahead and document the issue in case others run into it. Feel free to close it out if you don't want to have an open issue lying around without a repro case :)

kbrock:
Resolving classes at definition time can deadlock in production. AR holds a schema lock and a class-load lock. Loading one class that triggers loading another will hit both locks and hang.

- `belongs_to` and `has_many :through` now resolve ActiveHash classes at runtime, avoiding the deadlock
- Handle polymorphic `has_many :through` with `source_type` (fixes #334)

Note: `belongs_to_active_hash` is unchanged and still available for direct use.

Thank you, @alexgriff, for the reproducer.

### DIFF
--- a/lib/associations/associations.rb
+++ b/lib/associations/associations.rb
@@ -7,36 +7,48 @@ module ActiveHash
       end
 
       def has_many(association_id, scope = nil, **options, &extension)
+        super
+
         if options[:through]
           source_association_name = options[:source]&.to_s || association_id.to_s.singularize
 
-          through_klass = reflect_on_association(options[:through])&.klass
-          klass = through_klass&.reflect_on_association(source_association_name)&.klass
+          define_method(association_id) do
+            through_klass = self.class.reflect_on_association(options[:through])&.klass
+            source_klass = through_klass&.reflect_on_association(source_association_name)&.class_name&.safe_constantize
 
-          if klass && klass < ActiveHash::Base
-            define_method(association_id) do
-              join_models = send(options[:through])
-              join_models.flat_map do |join_model|
+            if source_klass && source_klass < ActiveHash::Base
+              send(options[:through]).flat_map do |join_model|
                 join_model.send(source_association_name)
               end.uniq
+            else
+              super()
             end
-
-            return
           end
         end
-
-        super
       end
 
       def belongs_to(name, scope = nil, **options)
         klass_name = options.key?(:class_name) ? options[:class_name] : name.to_s.camelize
-        klass = klass_name.safe_constantize
+        foreign_key = options[:foreign_key] || name.to_s.foreign_key
 
-        if klass && klass < ActiveHash::Base
-          options = { class_name: klass_name }.merge(options)
-          belongs_to_active_hash(name, options)
-        else
-          super
+        super
+
+        define_method(name) do
+          klass = klass_name.safe_constantize
+          if klass && klass < ActiveHash::Base
+            klass.send("find_by_#{klass.primary_key}", send(foreign_key))
+          else
+            super()
+          end
+        end
+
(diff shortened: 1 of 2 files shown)


kbrock:
update: removed class load
update: removed another class loader (total tangent, but solves a similar issue)

lib/associations/associations.rb
```diff
@@ -7,36 +7,63 @@ def self.extended(base)
...
-          if klass && klass < ActiveHash::Base
             define_method(association_id) do
-              join_models = send(options[:through])
-              join_models.flat_map do |join_model|
-                join_model.send(source_association_name)
-              end.uniq
+              klass = source_type.safe_constantize
+              if klass < ActiveHash::Base
+                ids = send(options[:through]).map { |jm| jm.send(source_foreign_key) }.compact.uniq
+                ids.flat_map { |id| klass.find_by_id(id) }.compact
```
flavorjones: This could potentially make a lot of queries. Could we instead do:

```suggestion
                klass.where(id: ids)
```

? If we need to keep the order we could do something like this to keep it to a single query:

```ruby
records = klass.where(id: ids).index_by(&:id)
ids.filter_map { |id| records[id] }
```

flavorjones (commented):
@kbrock Sorry for not reviewing this earlier. It looks good, I left one comment about whether using `:source_type` will generate N+1 queries, but I think we can just collapse that to a `where(id: ids)`.

lib/associations/associations.rb
kbrock: Thanks. This is accessing a Ruby hash, so I'm pretty sure it is reading from memory and basically does the same thing under the covers, but I like how the `where` clause reads, and it allows a user to use `order(:id)`.

kbrock:
update: fixed N+1 entry and ensure it is exposing a relation instead of an array.

@flavorjones thanks for the suggestion. Pretty sure I t was looking up in the local hash, but swapping with a where clause is much better to read. Thanks

flavorjones (approved):

PR #348, opened 2026-06-12
Fix nil FK queries, strict_where validation, field name conflict warnings

### LINKED ISSUE #184: GOTCHA - Cannot use the field name 'display' in ActiveHash as Rails adds it to Object class
I was attempting to use a field with name "display"; example:

```
class TestHash < ActiveHash::Base
  field :display, default: true
  add
  add display: false
end
```

However, ActiveHash defines method only if an existing instance method does not already exist, and Rails adds this method to either Object or Kernel. Ex:
```
TestHash.first.display
#<TestHash:0x00007f870c24a780> => nil
TestHash.last.display
#<TestHash:0x00007f870c249f38> => nil
```

While I totally understand the rationale here, I think it would be really helpful if you could spit out a warning or error if we attempt to define a method that is already defined. The result of the above issue was that most of my test suite failed and I had countless lines outputting the class information for individual ActiveHash records.

### LINKED ISSUE #190: where(undefined_key: nil) returns relation including all records
### Description

I think `where` with undefined key should raise an error because

- It's same as ActiveRecord behavior.
- We cannot recognize typo and get a surprising result.

### Code to reproduce

```ruby
class Records < ActiveHash::Base
  self.data = [
    { id: 1, name: 'foo' },
    { id: 2, name: 'bar' },
    { id: 3, name: 'baz' }
  ]
end

pp Records.where(undefined_key: nil).to_a
# [
#    #<Records:0x00007fadaf51a5e0 @attributes={:id=>1, :name=>"foo"}>,
#    #<Records:0x00007fadaf519988 @attributes={:id=>2, :name=>"bar"}>,
#   #<Records:0x00007fadaf519348 @attributes={:id=>3, :name=>"baz"}>
# ]
```

### Other informations
- This will be breaking change

### LINKED ISSUE #302: Accessing belongs_to association without a key creates a useless query
# Environment

* Ruby 3.2.2
* Rails 7.1.1
* ActiveHash 3.2.1

# Problem

When accessing a `belongs_to` association, if the related id is missing then a query for `id IS NULL` is made. This is different from the standard ActiveRecord behaviour and I think in 99% of cases a redundant query. Seems to be caused by the internal use of `find_by_id` with a blank argument: https://github.com/active-hash/active_hash/blob/master/lib/associations/associations.rb#L177

You can reproduce and observe with this script:

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem "rails", "7.1.1"
  gem "sqlite3"
  gem "active_hash", "3.2.1"
end

require "active_hash"
require "active_record"
require "logger"

ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: "db/development.sqlite3")
  create_table :authors, force: true
  create_table :magazines, force: true do |t|
    t.belongs_to :author, null: true
  end
end

class Author < ActiveRecord::Base
end

class Book < ActiveHash::Base

  fields :author_id

  belongs_to :author
end

class Magazine < ActiveRecord::Base
  belongs_to :author, optional: true
end

book = Book.new(author_id: nil)
magazine = Magazine.create!(author_id: nil)

puts "-" * 50
puts "Accessing ActiveHash association with NULL key"
book.author
puts "-" * 50
puts "Accessing ActiveRecord association with NULL key"
magazine.author
puts "-" * 50
```

# Potential Solution

Don't bother querying if the key is NULL, if there is some weird edge case where you would want this then perhaps the behaviour could be enabled with a config option or optional argument when defining the association.

sawirricardo:
## Summary

Three low-hanging fixes addressing open issues.

### Fixes #302 — belongs_to with nil foreign key makes useless query

When accessing a `belongs_to` association with a nil foreign key, ActiveHash was still calling `find_by_*` which triggers a DB query. ActiveRecord skips this query when the key is nil.

**Fix:** Early return `nil` when foreign key is nil in:
- `ActiveRecordExtensions#belongs_to`
- `belongs_to_active_hash`
- `Methods#belongs_to`
- `Methods#has_one`

No breaking change — nil key previously returned nil anyway after a useless query.

### Fixes #190 — `where(undefined_key: nil)` returns all records

`where()` with an undefined field silently returns all records because `read_attribute` returns nil for unknown keys, and nil == nil matches everything.

**Fix:** Opt-in `strict_where` class attribute. When enabled:
```ruby
  self.strict_where = true
  fields :name, :code
end

Country.where(typo_key: nil)
# => ArgumentError: Unknown field(s): typo_key. Valid fields are: id, name, code
```

Also validates against data keys for models that use undeclared fields via `data=`. Opt-in only — no breaking change.

### Fixes #184 — `field :display` silently returns nil

Rails adds `display` to Object. ActiveHash skips defining the method if one already exists, but gives no indication. Users get silent nil returns and confusing test failures.

**Fix:** `warn()` in `validate_field` when field name conflicts with an existing method:
```ruby
class TestHash < ActiveHash::Base
  field :display
  # => warning: ActiveHash: Field :display on TestHash conflicts with an existing method...
end
```

Additive warning only — no breaking change.

## Test Results

276 specs pass, 2 pre-existing sqlite3 adapter failures unrelated to these changes.

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -21,7 +21,7 @@ module ActiveHash
   end
 
   class Base
-    class_attribute :_data, :dirty, :default_attributes, :scopes
+    class_attribute :_data, :dirty, :default_attributes, :scopes, :strict_where
 
     if Object.const_defined?(:ActiveModel)
       extend ActiveModel::Naming
@@ -233,6 +233,13 @@ module ActiveHash
         field_name = field_name.to_sym
         if [:attributes].include?(field_name)
           raise ReservedFieldError.new("#{field_name} is a reserved field in ActiveHash.  Please use another name.")
+        end
+
+        if method_defined?(field_name) || private_method_defined?(field_name)
+          warn(
+            "ActiveHash: Field :#{field_name} on #{name} conflicts with an existing method. " \
+            "The field will not override the existing method, and accessing it may return unexpected results."
+          )
         end
       end
 
--- a/lib/active_hash/relation.rb
+++ b/lib/active_hash/relation.rb
@@ -66,6 +66,7 @@ module ActiveHash
     end
 
     def where!(conditions_hash, inverted = false)
+      validate_where_keys!(conditions_hash)
       self.conditions << Condition.new(conditions_hash)
       self
     end
@@ -195,6 +196,23 @@ module ActiveHash
     private
 
     attr_writer :conditions, :order_values, :klass, :all_records
+
+    def validate_where_keys!(conditions_hash)
+      return unless conditions_hash.is_a?(Hash)
+      return unless klass.respond_to?(:strict_where) && klass.strict_where
+
+      valid_fields = klass.field_names + [klass.primary_key.to_sym]
+      invalid_keys = conditions_hash.keys.map(&:to_sym) - valid_fields
+
+      if invalid_keys.any? && klass.respond_to?(:data) && klass.data.is_a?(Array)
+        data_keys = klass.data.flat_map(&:keys).map(&:to_sym).uniq
+        invalid_keys -= data_keys
+      end
+
+      if invalid_keys.any?
+        raise ArgumentError, "Unknown field(s): #{invalid_keys.map(&:to_s).join(', ')}. Valid fields are: #{valid_fields.map(&:to_s).join(', ')}"
+      end
+    end
 
     def apply_conditions(records, conditions)
       return records if conditions.blank?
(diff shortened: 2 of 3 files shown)


flavorjones:
@sawirricardo Thanks very much for opening this PR!

However, if this is three individual fixes, we would prefer it to be broken out into three separate pull requests, so each change can be reviewed and merged on its own.

Also, I would like to see tests added that clearly describe the problem being fixed. If you need any help figuring out to craft an appropriate test, I'm happy to help -- just drop a note here!

I appreciate your time -- thanks for understanding how we like to work. I hope you're able to make time to contribute in the future! ❤️

PR #349, opened 2026-06-26
Add exists? to Relation

michebble:
This PR adds exists? to `Relation` so that users can call the method after applying a scope to an ActiveHash model. 

This brings the behaviour inline with ActiveRecord.

### Problem
Calling `exists?` after apply scope/s results in NoMethodError being raised.
```
  scope :north_america, -> { where(continent: 'North America') }
  scope :south_america, -> { where(continent: 'South America') }

  self.data = [
    { id: 1, name: 'US', continent: 'North America' },
    { id: 2, name: 'Brazil', continent: 'South America' }
  ]
end
```
Without PR changes
```ruby
Country.exists? 1
# true
Country.north_america.exists? 1
# undefined method 'exists?' for an instance of ActiveHash::Relation (NoMethodError)
```
With PR changes
```ruby
Country.exists? 1
# true
Country.north_america.exists? 1
# true
```
#### initial comment

> ActiveRecord allows calling `exists?` on a relation, however ActiveHash currently only allows calling it on the base model. Adding the method to Relation provides functionality that those familiar with ActiveRecord expect.
> 
> I needed to monkey patch my app to get this behaviour and would appropriate it if you would consider including this in the next release. I would like to get rid of the monkey patch if possible

### DIFF
--- a/lib/active_hash/base.rb
+++ b/lib/active_hash/base.rb
@@ -103,20 +103,6 @@ module ActiveHash
         end
       end
 
-      def exists?(args = :none)
-        if args.respond_to?(:id)
-          record_index[args.id.to_s].present?
-        elsif !args
-          false
-        elsif args == :none
-          all.present?
-        elsif args.is_a?(Hash)
-          all.where(args).present?
-        else
-          all.where(id: args.to_s).present?
-        end
-      end
-
       def insert(record)
         @records ||= []
         record[:id] ||= next_id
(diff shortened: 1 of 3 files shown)


flavorjones:
@michebble We'd welcome a pull request for this, if you're up for it!

michebble:
Thank you @flavorjones . I have updated the comment to make it clear what this PR changes. 
Please let me know if there is anything else I can do to get the PR in good shape.

kbrock (commented):
This looks great.

curious, did you try delegating exist? to :records?

.e.g.:

```ruby
module ActiveHash
  class Relation
    # about line 8
    delegate :sample, :exists?, to: :records
  end
end
```

or updating one of the entries there.

try it out, and then ping if you have any issues.
This code is ok, but if you notice, we tried to avoid duplicating the method definitions

michebble:
I was hoping it would be that simple at first, but `records` is just an `Array`. After stewing on it a little longer, I realised that `Base` delegates a lot of methods to `Relation` (as `all`) so we could do the same with `exists?`

I believe this matches the spirit of your suggestion and the specs are still green. Thank you for the comment.

kbrock (approved):
This is great. thanks

